PR #20, opened 2020-04-06
Avoid creating multiple assessments when user navigates back.

johnpatrickmorgan:
This PR:

- Keeps track of the newly created `assessmentId` on the first page of assessments (either `CovidTestScreen` or `HealthWorkerExposureScreen`)

This avoids creating new assessments whenever the user navigates back to the first page. Instead the existing one will be patched.

### DIFF
--- a/src/features/assessment/CovidTestScreen.tsx
+++ b/src/features/assessment/CovidTestScreen.tsx
@@ -102,7 +102,7 @@ export default class CovidTestScreen extends Component<CovidProps, State> {
         if (assessmentId == null) {
             userService.addAssessment(assessment)
                 .then(response => {
-                    this.props.route.params.assessmentId = response.data.id
+                    this.props.navigation.setParams({assessmentId: response.data.id})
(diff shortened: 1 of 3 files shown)


abubuwe:
Is this PR still required? Feel as though some of the changes have since been duplicated and merged?

johnpatrickmorgan:
@abubuwe I'll investigate now if it's still required.

johnpatrickmorgan:
> Is this PR still required? Feel as though some of the changes have since been duplicated and merged?

@abubuwe To follow up, none of the recent changes affected this issue (other than the fact that there is now a third potential start screen- level of isolation), so this PR stands. However, I wanted to check that the assessments are correctly updated on the backend before merging this, and I'm still waiting for necessary permissions to do so. I will update here when I can.

johnpatrickmorgan:
I've now checked and the assessments are updated correctly on the backend, so happy to merge when newly approved.

PR #60, opened 2020-04-09
Check user ip against locale when creating account

the-kenneth:
The first time a user tries to create an account we will check their ip against the selected country in app. If these do not match we popup a modal asking them to confirm their country.

### DIFF
--- a/src/core/user/UserService.ts
+++ b/src/core/user/UserService.ts
@@ -230,7 +230,7 @@ export default class UserService extends ApiClientBase {
         return await AsyncStorageService.getUserCount();
     }
 
-    async setUserCountInAsyncStorage() {
+    async getStartupInfo() {
         const response = await this.client.get<StartupInfo>('/users/startup_info/');
         UserService.ipCountry = response.data.ip_country;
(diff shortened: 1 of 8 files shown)


src/core/user/UserService.ts
```diff
@@ -230,8 +231,9 @@ export default class UserService extends ApiClientBase {
     }
 
     async setUserCountInAsyncStorage() {
```
JoshElgar, line 233: Maybe this function should be renamed

src/features/register/CountryIpModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+            {label: i18n.t('united-kingdom'), value: GB_CODE}
+        ];
+
+        if (isAndroid) {
+            items.unshift({label: i18n.t('choose-one-of-these-options'), value: ""});
+        }
+
+        return (
+            <Modal
+                animationType="fade"
```
JoshElgar, line 56: We should make a modal component at some point!

src/core/user/UserService.ts
JoshElgar, line 233: Maybe this function should be renamed as it's no longer just setting user count

PR #63, opened 2020-04-09
Optionally hide buttons on consent screens

selvach:
Pass in viewOnly route props to all Privacy & Terms screens to hide the buttons.


src/features/register/us/NursesConsentUS.tsx
```diff
@@ -219,44 +224,53 @@ export class NursesConsentUSScreen extends Component<PropsType, TermsState> {
...
-                                and{" "}
-                                <ClickableText onPress={() => this.props.navigation.navigate('PrivacyPolicyUS')}>Privacy
-                                    Policy</ClickableText>.
-                            </RegularText>
-                        </Body>
-                    </ListItem>
-
+                    {
+                        !this.viewOnly && (
+                            <View>
```
JoshElgar, line 229: Can use <> </> instead of wrapping with a view. https://reactjs.org/docs/fragments.html#short-syntax

PR #71, opened 2020-04-14
Update healthcare logic

JoshElgar:
Mainly updates the HealthWorkerExposureScreen and the About Your Work screens to add a new question (In the last day have you interacted with any patients?), switch two choices for the healthcareStaffOptions and accordingly change the validation for following fields. 

Small changes:
Removed isCompact from dropdown field as it was serving no purpose aside from breaking styling. There is already an onlyPicker prop that can be used if you only need the picker component and not the styling around it.

Added some labels / required validation messages.


src/core/user/UserService.ts
```diff
@@ -204,7 +204,7 @@ export default class UserService extends ApiClientBase {
             return false
         } else {
             await AsyncStorageService.setIsHealthWorker(
-                (patientProfileResponse.data.healthcare_professional === "yes_does_treat")
+                (patientProfileResponse.data.healthcare_professional === "yes_does_interact")
```
abubuwe, line 207: we should keep the old option - some users will log-in with that value returned and should still see the screen,

src/features/assessment/HealthWorkerExposureScreen.tsx
```diff
@@ -67,46 +69,38 @@ export default class HealthWorkerExposureScreen extends Component<HealthWorkerEx
...
+            then: Yup.string().required(i18n.t("required-has-used-ppe-equipment")),
+        }),
+        ppeAvailabilityAlways: Yup.string().when(['interactedAnyPatients', 'hasUsedPPEEquipment'], {
+            is: (interactedAnyPatients, hasUsedPPEEquipment) => interactedAnyPatients === 'yes' && hasUsedPPEEquipment === 'always',
+            then: Yup.string().required(i18n.t("required-ppe-availability-always")),
         }),
         ppeAvailabilitySometimes: Yup.string().when('hasUsedPPEEquipment', {
-            is: 'sometimes',
-            then: Yup.string().required()
+            is: (interactedAnyPatients, hasUsedPPEEquipment) => interactedAnyPatients === 'yes' && hasUsedPPEEquipment === 'sometimes',
```
abubuwe, line 98: you expect two params here but your when only contains 1 form value?

src/features/assessment/HealthWorkerExposureScreen.tsx
```diff
@@ -67,46 +69,38 @@ export default class HealthWorkerExposureScreen extends Component<HealthWorkerEx
...
+            then: Yup.string().required(i18n.t("required-ppe-availability-sometimes"))
         }),
         ppeAvailabilityNever: Yup.string().when('hasUsedPPEEquipment', {
-            is: 'never',
-            then: Yup.string().required()
+            is: (interactedAnyPatients, hasUsedPPEEquipment) => interactedAnyPatients === 'yes' && hasUsedPPEEquipment === 'never',
```
abubuwe, line 102: you expect two params here but your when only contains 1 form value?

src/core/user/UserService.ts
JoshElgar: Good point

PR #78, opened 2020-04-14
Fix excessive whitespace at end of screens

abubuwe:
This PR removes the behavior prop from KeyboardAvoidingView. 

While the [docs](https://reactnative.dev/docs/keyboardavoidingview#behavior) specify that this prop is recommended for both iOS and Android, at present we only pass it for iOS (where the issue occurs) and it's removal prevents the additional padding from appearing without any obvious side effects that affect usability of the screen.

### DIFF
--- a/src/features/assessment/DescribeSymptomsScreen.tsx
+++ b/src/features/assessment/DescribeSymptomsScreen.tsx
@@ -206,7 +206,7 @@ export default class DescribeSymptomsScreen extends Component<SymptomProps, Stat
               >
                   {props => {
                       return (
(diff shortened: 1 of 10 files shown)


MikeD-ZOE:
The DescribeSymptoms screen has the last field of "Other symptoms" which is a text field. This pops up a keyboard. We've had a steady stream of reports that when trying to enter text on this field, it is hidden behind the keyboard, so users can't see what they've entered.

Similarly, the AboutYou collects height, weight and postcode via overlaid keyboard. Can a user scroll the page underneath the keyboard to make the Next Question button visible and tappable?

Have you tested both scenarios in both iOS and Android?

PR #79, opened 2020-04-14
Add extra screen prompting user to log for another.

selvach:
Adds an extra screen prompting user to log for another, if they haven't done so already.

### DIFF
--- a/src/features/multi-profile/ReportForOtherScreen.tsx
+++ b/src/features/multi-profile/ReportForOtherScreen.tsx
@@ -9,6 +9,7 @@ import {StackNavigationProp} from "@react-navigation/stack";
 import {profilesIcon} from "../../../assets";
 import {Text} from "native-base";
 import {getLocalThankYou} from "../Navigation";
+import UserService from "../../core/user/UserService";
 
 type RenderProps = {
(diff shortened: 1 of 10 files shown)


src/features/multi-profile/ReportForOtherScreen.tsx
```diff
@@ -44,7 +45,7 @@ export default class ReportForOtherScreen extends Component<RenderProps, {}> {
                                 Not right now? You can add additional profiles from the menu icon above at any time.
```
abubuwe, line 45: Is this true? I can't see it in the menu

selvach: Hmmm... probably. Not super obvious from designs. I'll add an item to the Drawer Menu.

selvach: Actually... lets park this. It might start breaking things if you can just jump straight to this page. Need to have a think!

abubuwe: 👍 I think it's okay for now in the SelectProfile screen as they're always shown that.

PR #87, opened 2020-04-15
Default smoker status to never -- the default option that exists.

MikeD-ZOE:
initial value of `smokerStatus` is `no`, an option that doesn't exist in the dropdown, so no default value is displayed to the user. The backend, however is sent the value of `never` in this case. This PR fixes the discrepancy between what the user sees on the screen (blank), and what is sent to the server (`never`).


JoshElgar:
Do you mind updating the data dictionary with a note to reflect this?

PR #88, opened 2020-04-15
Add avatarName as optional parameter to CreateProfile screen.

MikeD-ZOE:
Typescript is raising errors because the CreateProfile screen is getting `avatarName` from `this.props`, and it isn't defined as a param on ScreenListParams. This fixes that.


abubuwe:
This change has already been made in `develop`. Hence the merge conflicts I believe.

PR #96, opened 2020-04-15
Truncate navigation stack of future screens in profile creation/selection

MikeD-ZOE:
Attempts to fix the issues of users navigating backwards through profile creation pages, and solving the issue of revisiting patient screens to see them already populated by previous flows through the pages. This latter is a navigation issue in that if a page already exists on the navigation stack, then navigating to that screen redisplays that screen, rather than showing a new one. So the approach originally taken is correct, but it's important to preserve the stack before the Select profile page, which this PR attempts to do.

I've hit the limit of my TypeScript + generics knowledge, and patience, so I'll need help to refactor the `pushToCleanNavStack` and `cleanNavigationStack` methods into plain utility functions. So I've duplicated the code in SelectProfileScreen and ConsentForOtherScreen, and there's a few `: any` floating around.

### DIFF
--- a/src/features/multi-profile/ConsentForOtherScreen.tsx
+++ b/src/features/multi-profile/ConsentForOtherScreen.tsx
@@ -8,7 +8,7 @@ import {colors} from "../../../theme"
 import UserService from "../../core/user/UserService";
 import {StackNavigationProp} from "@react-navigation/stack";
 import {ConsentType, ScreenParamList} from "../ScreenParamList";
-import {RouteProp} from "@react-navigation/native";
+import {CommonActions, RouteProp} from "@react-navigation/native";
 import {PatientInfosRequest} from "../../core/user/dto/UserAPIContracts";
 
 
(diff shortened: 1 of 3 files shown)


src/features/multi-profile/ConsentForOtherScreen.tsx
```diff
@@ -65,13 +65,44 @@ export default class ConsentForOtherScreen extends Component<RenderProps, Consen
...
+        let foundRoute = false;
+        const newStack = navState.routes.filter(route => {
+            if (foundRoute) {
+                return false;
+            }
+            if (route.name === "SelectProfile") {
+                foundRoute = true;
+            }
+            return true;
+        })
```
JoshElgar, line 81: I don't understand this, could you explain? Where is foundRoute being used?

MikeD-ZOE: foundFoute is used inside the anonymous function. To allow me to take an array, and just return an array that contains all the routes upto-and-including "SelectProfile" route. Perhaps using a `while` instead of `Array.filter()` is a more readable approach?

JoshElgar: But nothing about this function suggests it would stop once it's 'foundRoute'?

JoshElgar: I see now! While might be more readable :)

the-kenneth:
Ive updated this now and tested it. Someone else to check it does have the intended behaviour would be great though

selvach:
Tested! LGTM! Thank you!

MikeD-ZOE:
I've tested Kenneth's changes, and satisfied that they fix the problem. Since I raised this PR, I can't approve it, but a solid 👍 from me.

PR #119, opened 2020-04-16
Add back button to Patient Header

MikeD-ZOE:
Screenshot 2020-04-16 at 15 30 37


selvach:
Added some styling to the icon!

PR #139, opened 2020-04-22
SWEDEN: hide screens: multiple patients and personal info

MikeD-ZOE:
This PR handles the screen-flow changes to support Sweden.

It adds country-level configuration to set "feature flags", and we use those feature flags to decide whether to show a feature or skip it. There's a default configuration, and country-specific configuration overrides the default. That way the country-level configuration just needs to specify the difference in configuration from the default.

Two feature-flag configurations:

* `enableMultiplePatients` -- for enabling multiple patient support (so we can turn it off for Sweden)
* `enablePersonalInformation` -- for enabling the Optional Information screen which collects personal identifying information (so we can skip it in Sweden)

There's a Navigator object, which I'm using to capture all the logic for determining what the next screen is, and navigating to it. The SplashScreen passes the navigation object to it, so you don't need to pass it on each call. Then we have methods to support both the "we're finished, go somewhere interesting", and "let's start a new process". This handles the three cases specific to Sweden:

* On the WelcomeRepeat screen starting a New report -- the decision whether to show SelectProfile or start report with the default patient. And then, starting a report, whether to go straight into the assessment, or start with collecting patient details
* On the registration page, deciding whether to show the OptionalInfo screen, or skip over it to the first PatientScreen (YourStudy, or YourWork)
* On the end of the assessment, deciding whether to show the "ReportForOthers" screen, or the locale-specific ThankYou page. I've also turned that from a `navigation.replace` to a `navigation.navigate`, I didn't see a reason why that was `replace`.


src/features/Navigation.ts
```diff
@@ -1,19 +1,186 @@
...
+    }
+
+    setNavigation(navigation: NavigationType) {
+        this.navigation = navigation;
+    }
+
+    // Workaround for Expo save/refresh nixing the navigation.
+    resetNavigation(navigation: NavigationType) {
+        if (!this.navigation) {
+            console.log("[ROUTE] Resetting navgation");
```
JoshElgar, line 32: console log

src/features/Navigation.ts
```diff
@@ -1,19 +1,186 @@
...
+            this.gotoScreen(this.getThankYouScreenName());
+        }
+    }
+
+    async gotoNextScreen(screenName: ScreenName, params: RouteParamsType) {
+        if (ScreenFlow[screenName]) {
+            ScreenFlow[screenName](params);
+        } else {
+            // We don't have nextScreen logic for this page. Explain loudly.
+            console.error("[ROUTE] no next route found for:", screenName);
```
JoshElgar, line 108: console

src/features/Navigation.ts
MikeD-ZOE: The resetNavigation is a workaround for developing with Expo. When you make a change to the page you are on, we lose the navigation reference. So reset is called to repopulate the navigation is missing. This console message then only displays in a dev environment to remind me we need a better way of repairing this

src/features/Navigation.ts
MikeD-ZOE: If code reaches here it's a genuine development error and needs to be fixes. This will trigger a red error screen in development.

PR #140, opened 2020-04-22
SWEDEN: Sv support and WelcomeScreen unification

the-kenneth:
This adds sv-SE and merges the welcome screens into one.

Im not completely happy with the triple ifs around to check which of the 3 locals you are currently in. But perhaps with Mike's upcoming locale config this could be cleaner?


src/features/ScreenParamList.tsx
```diff
@@ -11,10 +11,9 @@ export type ScreenParamList = {
 
     // Welcome screens
     Welcome: undefined;
-    WelcomeUS: undefined;
-    Welcome2US: undefined;
+    Welcome2: undefined;
+    WelcomeSV: undefined;
```
JoshElgar, line 15: Do we have a separate welcome screen for sweden?

src/features/register/WelcomeRepeatScreen.tsx
```diff
@@ -51,12 +49,36 @@ export class WelcomeRepeatScreen extends Component<PropsType, WelcomeRepeatScree
         this.props.navigation.navigate('SelectProfile');
     };
 
+    navigateToPrivacyPolicy = () => {
+        if (isUSLocale()) {
+            this.props.navigation.navigate('PrivacyPolicyUS', {viewOnly: true})
+        } else {
+            this.props.navigation.navigate('PrivacyPolicyUK', {viewOnly: true})
```
JoshElgar, line 56: My PR will add privacy policy sv too.

src/features/ScreenParamList.tsx
the-kenneth: No we dont, I'll remove that

the-kenneth:
OK added in final UI changes, ready for review.

JoshElgar:
Re: my previous comment on userCountry, we should document this in data dictionary I suppose

assets/lang/sv-SE.json
```diff
@@ -327,7 +327,7 @@
     "create-account": "Skapa konto"
   },
   "total-people-contributing": "%{total} personer bidrar",
-  "join-total-people-contributing": "Bli en av de %{total} personerna i världen som bidrar",
+  "join-total-people-contributing": "Anslut dig till %{total} andra runt om i världen som redan bidrar",
```
MikeD-ZOE, line 330: I think this translation string has been superceded, and what's in `develop` now is correct.

src/core/user/UserService.ts
```diff
@@ -27,7 +27,7 @@ const PATIENT_VERSION = '1.3.0';    // TODO: Wire this to something automatic.
 const MAX_DISPLAY_REPORT_FOR_OTHER_PROMPT = 3
 
 export default class UserService extends ApiClientBase {
-    public static userCountry = 'US';
+    public static userCountry = 'en-US';
```
MikeD-ZOE, line 30: I'm starting to get a bad feeling about using locale as a county.

As I understand it, a locale (for localisation) has two parts:
* A language
* A region/culture

I think country code should be the latter (the two character country code). Because we use that to switch between country specific questions. People in the UK get questions the NHS want answers to. The language doesn't affect which screens or questions the user gets. The language should just pull in the correct language text for that region.

I'm using the country as a key to getting the correct country-specific configuration. So I have SE.json, GB.json and US.json -- the above change breaks this, I'm probably going to need to trim off the language part to get the key I need -- because `sv-SE` and `en-SE` should use the same configuration file because the country is Sweden.

This may be just naming, perhaps what we need is instead of adding `en-` to the `userCountry`, create a separate property: `userLocale` -- which would correctly be `en-US`?

the-kenneth: Yeh, this is the topic I wanted to have a chat about. I agree with you. We need to split the 2 properly now we are introducing more languages

assets/lang/sv-SE.json
the-kenneth: Done

PR #143, opened 2020-04-23
SWEDEN: Hide Cohorts and Race/Ethnicity

MikeD-ZOE:
Both using a country-specific feature flag.

* Skip over the cohort screen (Population Studies, Your Clinical Study) in the StartPatient screenflow if the `enableCohorts` flag is false
* Add `showEthnicityQuestion` and `showRaceQuestion` feature flags, and disable both for Sweden.

I'm not happy with the async dependency on get country to get the  config file. It just means having to use component state to populate the values when they are retrieved. Otherwise we hit a race-condition, because render doesn't wait for componentDidMount'ed promise.

Also, props.navigation might be stateful, even though it just looks like a group of methods. So maybe passing it in on the SplashScreen is insufficient, and we need to pass it in on each component. running `Navigator.replaceScreen` had issues without passing in the more recent props.navigation.

### DIFF
--- a/src/features/patient/AboutYouScreen.tsx
+++ b/src/features/patient/AboutYouScreen.tsx
@@ -22,9 +22,6 @@ import {CheckboxItem, CheckboxList} from "../../components/Checkbox";
 import {cloneDeep} from 'lodash';
 
 
-
-const PICKER_WIDTH = (Platform.OS === 'ios') ? undefined : '100%';
-
 const initialFormValues = {
     yearOfBirth: "",
     sex: "",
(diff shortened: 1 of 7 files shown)


src/features/patient/AboutYouScreen.tsx
```diff
@@ -90,21 +87,35 @@ type AboutYouProps = {
...
 
 
 export default class AboutYouScreen extends Component<AboutYouProps, State> {
-
     constructor(props: AboutYouProps) {
         super(props);
         this.state = initialState;
     }
 
+    async componentDidMount() {
```
jamesbrooks94, line 110: Why not migrate this to a arrow function 
https://dev.to/trentyang/replace-lifecycle-with-hooks-in-react-3d4n#componentdidmount

You can replace componentDidMount with useEffect, then set the state based on the response!

MikeD-ZOE: Thanks for the comment and your participation!

I've tried out React hooks recently, I liked what I saw. But internally we've standardised on the class-based way for React-Native, to keep the pre-knowledge-barrier low so our current team of developers can help out too. With that we've had Python developers and native mobile app developers able to jump in and build functionality in a fairly short period of time.

You are right, a global Context object feels like the next refactoring step, for the very reasons you lay out. :-)

jamesbrooks94: OK no problem, good to know for future PR's :)

PR #147, opened 2020-04-23
chore(Types): Improve types on YourWorkScreen and migrate Passwords to arrow fns

jamesbrooks94:


src/CovidApp.tsx
```diff
@@ -1,142 +1,274 @@
...
-                        width: Dimensions.get("screen").width,
-                    }}>
-                        <Drawer.Screen name="Main" component={this.mainNavStack}/>
-                    </Drawer.Navigator>
-                </NavigationContainer>
-            </Root>
-        );
-    }
+  render() {
+    if (!this.state.fontLoaded) return <View />;
```
JoshElgar, line 73: Can return null or </> in general if you don't want anything to render

jamesbrooks94: Missed this, was part of a refactor :)

Issue #91, opened 2020-04-15
Progress loader styling issues


livalana:
Progress loader should be 2px thick (in build it's 5px). The loader colour should be #A10056 (same as links colour)

![image](https://user-images.githubusercontent.com/12391240/79329012-e830bf80-7f0e-11ea-80e7-fb655ee1950a.png)

PR #149, opened 2020-04-23
fix(Progress Loader): Fix progress loader height

jamesbrooks94:
Fixes https://github.com/zoe/covid-tracker-react-native/issues/91


jamesbrooks94:
> Thanks James.
> #91 also mentions the color being the wrong one, is that something you could append to this PR maybe?

I missed that, updated :)

Issue #34, opened 2020-04-07
Signup / Login not working with password managers


selvach:

louiechristie:
I did manage to use 1Password to sign up on the production app, but with a couple of small issues.

Environment:
App: COVID Symptom Study version 1.9 (Production App Store App)
Password manager: 1Password 7.5.2
OS: iOS 13.4.1
Browser: Safari

Issues:

1. In 1Password when clicking Create Login, the name of the app, and the website were not auto populated.
1. When logging in to 1Password with password (Face ID works ok) intermittently the keyboard overlaid the password box, blocking view. But I don't think this is an issue with COVID app, think that is an issue with Safari, iOS or 1Password.

PR #160, opened 2020-04-24
fix(Password managers): Attempt to fix password managers on iOS

jamesbrooks94:
# Description

Attempt to fix password managers not saving credentials on iOS. Struggling to get this working in a non-production environment. This could fix https://github.com/zoe/covid-tracker-react-native/issues/34

- [ ] Android - Emulator
- [X] iOS - Emulator
- [X] iOS - Physical device

## Screenshots

N/A

This may need further testing and a follow up PR. Documentation followed: 
https://docs.expo.io/workflow/configuration/?redirected#ios
https://gist.github.com/amcvitty/42cbe072184fe72485ad17cd7120bb89#add-the-associated-domains-entitlement

```js
/*
      An array that contains Associated Domains for the standalone app. See apple's docs for config: https://developer.apple.com/documentation/uikit/core_app/allowing_apps_and_websites_to_link_to_your_content/enabling_universal_links

      Entries must follow the format "applinks:<fully qualified domain>[:port number]". See Apple's docs for details -> https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_associated-domains

      ExpoKit: use Xcode to set this.
    */
    "associatedDomains": ARRAY,
```
### This may also need to be done:

https://gist.github.com/amcvitty/42cbe072184fe72485ad17cd7120bb89#add-the-apple-app-site-association-file-to-your-website

### DIFF
--- a/app.json
+++ b/app.json
@@ -24,9 +24,10 @@
       "**/*"
     ],
     "ios": {
-      "supportsTablet": true,
+      "associatedDomains": ["webcredentials:covid-zoe.joinzoe.com"],
+      "buildNumber": "0.0.27",
       "bundleIdentifier": "com.joinzoe.covid-zoe",
-      "buildNumber": "0.0.27"
+      "supportsTablet": true
     },
     "android": {
       "package": "com.joinzoe.covid_zoe",


selvach:
We are going to close this, as we don't really have the resource to investigate the fix and get comfortable with deploying it. Maybe we'll re-visit this in quieter times....

PR #167, opened 2020-04-24
SWEDEN consent flow 4

JoshElgar:
# Description
The new swedish consent flow:
- swedish consent
- privacy policy
- refactored consent screen.
- new document versions

- [ ] Android - Emulator
- [X] Android - Physical device
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots
Screenshot_20200424-180554

Screenshot_20200424-180546

Look into using a markdown renderer if we feel that updating the legal docs is becoming a large time investment - the 2-3 legal components I have made are one-offs in that they're built for fast copy pasting and condensing large swathes of text.

### DIFF
--- a/src/features/register/ConsentScreen.tsx
+++ b/src/features/register/ConsentScreen.tsx
@@ -83,7 +83,7 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
     };
 
     handleSVAgreeClicked = async () => {
-        await this.userService.setConsentSigned("SV", consentVersionSV, privacyPolicyVersionSV);
+        await this.userService.setConsentSigned("SE", consentVersionSV, privacyPolicyVersionSV);
         this.props.navigation.navigate('Register')
     };
 
--- a/src/features/register/ConsentScreen.tsx
+++ b/src/features/register/ConsentScreen.tsx
@@ -94,7 +94,6 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
 
     renderUSConsent = () => {
         return (
-            isUSCountry() ? (
                     <ScrollView>
                         <RegularText>
                             If you are in an existing research or clinical study (e.g. Nurses’ Health Studies) and you want your data to be shared with investigators on that study, <ClickableText
@@ -170,7 +169,6 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
                         }
 
                     </ScrollView>
-            ) : (<View/>)
         )
     };
 
--- a/src/features/register/ConsentScreen.tsx
+++ b/src/features/register/ConsentScreen.tsx
@@ -15,6 +15,7 @@ import {
 } from "./constants";
 import {RouteProp} from "@react-navigation/native";
 import { HeaderText, SimpleTextBlock } from "./LegalComponents";
+import i18n from "../../locale/i18n";
 
 
 type PropsType = {
@@ -50,24 +51,15 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
 
     viewOnly = this.props.route.params.viewOnly;
 
-    handleSwedenParticipateChange = () => {
-        this.setState({swedenParticipateChecked: !this.state.swedenParticipateChecked})
-    };
-
-    handleSwedenProcessingChange = () => {
-        this.setState({swedenProcessingChecked: !this.state.swedenProcessingChecked})
-    };
-
-    handleSwedenAgreeZoe = () => {
-        this.setState({swedenAgreeZoe: !this.state.swedenAgreeZoe})
-    };
-
-    handleProcessingChange = () => {
-        this.setState({processingChecked: !this.state.processingChecked})
-    };
-
-    handleTermsOfUseChange = () => {
-        this.setState({termsOfUseChecked: !this.state.termsOfUseChecked})
+    handleToggle = (label: string) => {
+        if (label in this.state) {
+            // @ts-ignore
+            let val: any = this.state[label];
+            this.setState((prevState) => ({
+                ...prevState,
+                [label]: !val
+            }));
+        }
     };
 
     handleUSAgreeClicked = async () => {
@@ -100,13 +92,13 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
         return false;
     };
 
-    renderConsentAgreeButton = (props?: any) => {
+    renderConsentAgreeButton = () => {
         return (
             <BrandedButton
               style={styles.button}
               enable={this.canAgree()}
               hideLoading={true}
-              onPress={this.handleAgreeClicked} {...props}>I agree</BrandedButton>
+              onPress={this.handleAgreeClicked}>{i18n.t('legal.i-agree')}</BrandedButton>
         )
     };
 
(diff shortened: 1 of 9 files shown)


src/features/register/ConsentScreen.tsx
```diff
@@ -16,6 +23,9 @@ type PropsType = {
 }
 
 interface TermsState {
+    swedenParticipateChecked: boolean,
+    swedenProcessingChecked: boolean,
+    swedenAgreeZoe: boolean,
```
the-kenneth, line 28: Cant we reuse existing fields and only create new ones for anything new? I think reusing will be of benefit when we add more countries

the-kenneth: I think we should make it more generic and get rid of the sweden specific functions/fields

src/features/register/sv/PrivacyPolicySVScreen.tsx
```diff
@@ -0,0 +1,175 @@
...
+                    <RegularText>
+                        Vår adress: 164 Westminster Bridge Road, London SE1 7RW, Storbritannien{"\n\n"}
+                        Dataskyddsansvarig: dpo@joinzoe.com
+                        {"\n"}
+                    </RegularText>
+
+                </ScrollView>
+
+                {
+                    !this.viewOnly && <BrandedButton style={styles.button} onPress={() => this.props.navigation.goBack()}>Back</BrandedButton>
```
abubuwe, line 150: Need i18n - Swedish is Tillbaka

src/features/register/ConsentScreen.tsx
```diff
@@ -53,9 +82,35 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
...
+        return false;
+    };
+
+    renderConsentAgreeButton = (props?: any) => {
+        return (
+            <BrandedButton
+              style={styles.button}
+              enable={this.canAgree()}
+              hideLoading={true}
+              onPress={this.handleAgreeClicked} {...props}>I agree</BrandedButton>
```
abubuwe, line 109: Needs i18n - Swedish is Jag godkänner

src/features/register/ConsentScreen.tsx
```diff
@@ -53,9 +82,35 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
...
+    };
+
+    canAgree = () => {
+        if (isUSCountry()) return this.state.processingChecked && this.state.termsOfUseChecked;
+        if (isGBCountry()) return true;
+        if (isSECountry()) return this.state.swedenParticipateChecked && this.state.swedenProcessingChecked && this.state.swedenAgreeZoe;
+        return false;
+    };
+
+    renderConsentAgreeButton = (props?: any) => {
```
abubuwe, line 103: Does this really need to be it's own function? Could just be used directly, the same way the back button is implemented in the Privacy Policy screens?

src/features/register/ConsentScreen.tsx
```diff
@@ -187,13 +242,161 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
...
+    };
+
+    renderSwedishConsent = () => {
+        return (
+            <View style={styles.rootContainer}>
+                    <ScrollView>
+                        <RegularText>
+                            Vi vill fråga dig om du vill delta i ett forskningsprojekt som handlar om covid-19. I det här dokumentet får du information om projektet och om vad det innebär att delta.
+                            {"\n\n"}
+                            Projektets titel är ”Nationellt initiativ för att via en app i realtid kartlägga samhällspridningen av covid-19 i Sverige samt riskfaktorer för att drabbas av en allvarlig sjukdomsbild vid covid-19”, och forskningen är godkänd av Etikprövningsmyndigheten i Sverige (DNR 2020-XXX).
```
abubuwe, line 259:
```suggestion
                            Projektets titel är ”Nationellt initiativ för att via en app i realtid kartlägga samhällspridningen av covid-19 i Sverige samt riskfaktorer för att drabbas av en allvarlig sjukdomsbild vid covid-19”, och forskningen är godkänd av Etikprövningsmyndigheten i Sverige (DNR 2020-01803).
```

src/features/register/sv/PrivacyPolicySVScreen.tsx
jamesbrooks94: Can this be moved out to a language file?

src/features/register/ConsentScreen.tsx
JoshElgar: I didn't re-use because I didn't want us to need to worry about if someone goes to the US consent - signs it, goes back, changes country, then skips past that consent checkbox, especially as the consents/policies are actually different. May not be a problem depending on how the navigation replaces / refreshes the consent page.

src/features/register/ConsentScreen.tsx
JoshElgar: Totally could be, but why? This way I only have to update it in one place rather than three.

src/features/register/ConsentScreen.tsx
JoshElgar: Good spots

src/features/register/ConsentScreen.tsx
abubuwe: it shouldn't need to be repeated 3 times? If the page stays the same and just the copy/check boxes are changing?

src/features/register/ConsentScreen.tsx
```diff
@@ -53,9 +82,35 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
...
+              style={styles.button}
+              enable={this.canAgree()}
+              hideLoading={true}
+              onPress={this.handleAgreeClicked} {...props}>I agree</BrandedButton>
+        )
+    };
+
+    renderUSConsent = () => {
         return (
             isUSCountry() ? (
```
abubuwe, line 115: this check isn't necessary as you make it below in what is now line 389.

src/features/register/ConsentScreen.tsx
```diff
@@ -187,13 +242,161 @@ export class ConsentScreen extends Component<PropsType, TermsState> {
...
+                                        </Body>
+                                    </ListItem>
+
+                                </View>
+                            )
+                        }
+
+                    </ScrollView>
+
+                    {
```
abubuwe, line 379: NTH: This could be refactored to only appear once rather than the 3 times it currently is repeated.

src/features/register/ConsentScreen.tsx
JoshElgar: Oh, I think I see what you're getting at - does my latest commit solve that? Also, there's a far larger chunk of re-used code in the checkbox section that should be cut down

src/features/register/ConsentScreen.tsx
JoshElgar: Refactored the views, left scrollview for now as we still need some wrapper component around all the text, so not really much point removing it just to replace with <> or <View>

src/features/register/ConsentScreen.tsx
JoshElgar: I refactored the functions so we can re-use the toggle functionality for checkboxes, but kept the sweden fields, is that alright?

PR #185, opened 2020-04-26
Fix dropdown translation bugs

selvach:
# Description

Fixes dropdown option translations. These options are were previously defined in `YourWorkScree/helpers.ts` and imported before i18n.locale is initialised  and so the default `en` is used. Super odd though, as I would assume this would be called lazily and my ReactJS is not strong enough to figure out a better solution. 

```
Logging:

In helpers.ts
Locale: en
Running application on Pixel 2 XL.
Call to getConfig() in GetPatient
Locale: sv-SE
```

This only happens when you first launch the app. If you do an Expo live-reload, it no longer presents itself. Again ... not sure why? 

What do you think @jamesbrooks94 ?

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

### DIFF
--- a/src/features/patient/YourWorkScreen/helpers.ts
+++ b/src/features/patient/YourWorkScreen/helpers.ts
@@ -1,7 +1,6 @@
 import { RouteProp } from '@react-navigation/native';
 import { StackNavigationProp } from '@react-navigation/stack';
 
-import { isAndroid } from '../../../components/Screen';
 import {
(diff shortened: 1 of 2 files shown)


jamesbrooks94:
I’ll double check it in the morning but looks good, might be worth investigating why this happens, is it a race condition?

PR #187, opened 2020-04-27
Remove async/await from getting country config.

MikeD-ZOE:
# Description

getting country config depends on getting the country value using asyncStorage. This PR removes that dependency, and hooks into setting the countrycode on UserService instead. Thus we can get the config file synchronously, which will help dealing with config flags needed synchronously before React render.

- [ ] Android - Emulator
- [X] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

### DIFF
--- a/src/core/Config.ts
+++ b/src/core/Config.ts
@@ -5,12 +5,16 @@ import DefaultConfig from './config/default.json';
 
 export type ConfigType = {
   country: string;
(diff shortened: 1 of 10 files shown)


MikeD-ZOE:
Since this PR has been merged in as part fo PR 186, I can close this.

PR #199, opened 2020-04-29
Add screen to ask back dated profile questions

the-kenneth:
# Description

Adds a screen before assessment that will ask the user any profile questions that may have been added to the app after they have already signed up.

Ive moved the blood pressure medication here and added the race/ethnicity question. This involves moving those questions out into their own components and using them in their original screens as well as the new one.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

It seems formik validation for the race question may have gotten broken before this, but it looks like its not working atm. I couldn't see the issue but will look again.

### DIFF
--- a/assets/lang/en.json
+++ b/assets/lang/en.json
@@ -93,7 +93,6 @@
     "back": "Back"
   },
 
-
   "next-question": "Next question",
   "completed": "Done",
   "something-went-wrong": "Something went wrong, please try again later",
@@ -108,6 +107,9 @@
     "worked-school-clinic": "School clinic",
     "worked-other-facility": "Other health care facility"
   },
+
+  "back-date-profile-title": "Some additional questions about you",
+  "update-profile": "Update profile",
 
   "title-about-work": "About your work",
   "label-cohort": "Are you part of one of these population studies or organisations?",
--- a/src/core/user/UserService.ts
+++ b/src/core/user/UserService.ts
@@ -160,18 +160,19 @@ export default class UserService extends ApiClientBase {
 
   public async updatePatientState(patientState: PatientStateType, patient: PatientInfosRequest) {
     // Calculate the flags based on patient info
+    const hasRaceAnswer = patient.race.length > 0;
     const isFemale = patient.gender == 0;
     const isHealthWorker =
       ['yes_does_treat', 'yes_does_interact'].includes(patient.healthcare_professional) ||
       patient.is_carer_for_community;
     const hasBloodPressureAnswer =
-      patient.takes_any_blood_pressure_medications === true || patient.takes_any_blood_pressure_medications === false;
+      patient.takes_any_blood_pressure_medications || !patient.takes_any_blood_pressure_medications;
     const hasCompletePatientDetails =
       // They've done at least one page of the patient flow. That's a start.
       !!patient.profile_attributes_updated_at &&
       // If they've completed the last page, heart disease will either be true or false
       // and not null. (or any nullable field on the last page)
-      (patient.has_heart_disease === true || patient.has_heart_disease === false);
+      (patient.has_heart_disease || !patient.has_heart_disease);
 
     let patientName = patient.name;
     if (!patientName || (!patient.reported_by_another && patientName === 'Me')) {
@@ -203,6 +204,7 @@ export default class UserService extends ApiClientBase {
       profile,
       isFemale,
       isHealthWorker,
+      hasRaceAnswer,
       hasBloodPressureAnswer,
       hasCompletePatientDetails,
       isReportedByAnother,
--- a/src/core/user/UserService.ts
+++ b/src/core/user/UserService.ts
@@ -165,14 +165,13 @@ export default class UserService extends ApiClientBase {
     const isHealthWorker =
       ['yes_does_treat', 'yes_does_interact'].includes(patient.healthcare_professional) ||
       patient.is_carer_for_community;
-    const hasBloodPressureAnswer =
-      patient.takes_any_blood_pressure_medications || !patient.takes_any_blood_pressure_medications;
+    const hasBloodPressureAnswer = patient.takes_any_blood_pressure_medications != null
     const hasCompletePatientDetails =
       // They've done at least one page of the patient flow. That's a start.
       !!patient.profile_attributes_updated_at &&
       // If they've completed the last page, heart disease will either be true or false
       // and not null. (or any nullable field on the last page)
-      (patient.has_heart_disease || !patient.has_heart_disease);
+      (patient.has_heart_disease != null);
 
     let patientName = patient.name;
     if (!patientName || (!patient.reported_by_another && patientName === 'Me')) {
--- a/src/features/assessment/ProfileBackDateScreen.tsx
+++ b/src/features/assessment/ProfileBackDateScreen.tsx
@@ -1,0 +1,235 @@
+import { RouteProp } from '@react-navigation/native';
+import { StackNavigationProp } from '@react-navigation/stack';
+import { Formik, FormikProps } from 'formik';
+import { Form, Item, Label, Text } from 'native-base';
+import React, { Component } from 'react';
+import { Platform } from 'react-native';
+import * as Yup from 'yup';
+
+import DropdownField from '../../components/DropdownField';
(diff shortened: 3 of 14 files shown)


src/core/user/UserService.ts
```diff
@@ -160,18 +160,19 @@ export default class UserService extends ApiClientBase {
...
   public async updatePatientState(patientState: PatientStateType, patient: PatientInfosRequest) {
     // Calculate the flags based on patient info
+    const hasRaceAnswer = patient.race.length > 0;
     const isFemale = patient.gender == 0;
     const isHealthWorker =
       ['yes_does_treat', 'yes_does_interact'].includes(patient.healthcare_professional) ||
       patient.is_carer_for_community;
     const hasBloodPressureAnswer =
-      patient.takes_any_blood_pressure_medications === true || patient.takes_any_blood_pressure_medications === false;
+      patient.takes_any_blood_pressure_medications || !patient.takes_any_blood_pressure_medications;
```
abubuwe, line 169: I believe this is no longer correct - hasBloodPressureAnswer will be set to true, even if backend value is null?

src/core/user/UserService.ts
```diff
@@ -160,18 +160,19 @@ export default class UserService extends ApiClientBase {
...
     const hasCompletePatientDetails =
       // They've done at least one page of the patient flow. That's a start.
       !!patient.profile_attributes_updated_at &&
       // If they've completed the last page, heart disease will either be true or false
       // and not null. (or any nullable field on the last page)
-      (patient.has_heart_disease === true || patient.has_heart_disease === false);
+      (patient.has_heart_disease || !patient.has_heart_disease);
```
abubuwe, line 175: I believe this is no longer correct - hasBloodPressureAnswer will be set to true, even if backend value is null

src/core/user/UserService.ts
MikeD-ZOE: The original line will result in one of three values: True, False, or null/undefined;
The changed line will result in one of two values: True or False, because the second last part takes a falsy value (false, '', 0, null, undefined) and coerce it into false. So we lose the difference between false, and falsy.

In practice that means before we could tell the difference between a patient answering the question no, or not answering. With the change, we no longer know.

This is one of the quirks of Javascript. Truthiness and Falsiness are sometimes different to True and False.

src/features/assessment/ProfileBackDateScreen.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  doesChemiotherapy: 'no',
+  takesImmunosuppressants: 'no',
+  takesCorticosteroids: 'no',
+  takesAspirin: 'no',
+  takesBloodPressureMedications: 'no', // pril
+  takesAnyBloodPressureMedications: 'no',
+  takesBloodPressureMedicationsSartan: 'no',
+
+  limitedActivity: 'no',
```
MikeD-ZOE, line 67: I was expecting this set of initial values to just be the list of fields on the page, instead of all the AboutYou fields.

Later on I'd like us to solve how to populate this by asking `RaceEthnicityQuestion` and `BloodPressureMedicationQuestion` to give us their initial form values.

src/features/assessment/ProfileBackDateScreen.tsx
```diff
@@ -0,0 +1,235 @@
...
+      then: Yup.array<string>().min(1, i18n.t('please-select-race')),
+    }),
+    raceOther: Yup.string().when('race', {
+      is: (val: string[]) => val.includes('other'),
+      then: Yup.string().required(),
+    }),
+    ethnicity: Yup.string().when([], {
+      is: () => isUSCountry(),
+      then: Yup.string().required(),
+    }),
```
MikeD-ZOE, line 123: And here, as future next steps, how to get `RaceEthnicityQuestion` and `BloodPressureMedicationQuestion` to give us their validation rules.

src/features/assessment/ProfileBackDateScreen.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  handleProfileUpdate(formData: BackData) {
+    const { currentPatient } = this.props.route.params;
+    const patientId = currentPatient.patientId;
+
+    const userService = new UserService();
+
+    let infos: Partial<PatientInfosRequest> = {};
+
+    if (formData.takesAnyBloodPressureMedications) {
```
MikeD-ZOE, line 140: Again, another future refactoring: How to get `RaceEthnicityQuestion` and `BloodPressureMedicationQuestion` to generate their own patient docs, so here we combine the documents from both field components.

Once we figure out these next steps, then we can have a Form Screen that takes an array of "Fields", and renders the form by asking the fields for: the initial values, field validation rules, render itself, assembles the patient document by asking each field for their sub-documents.

Then we can rename the app SurveyMonkey.

src/features/assessment/ProfileBackDateScreen.tsx
```diff
@@ -0,0 +1,235 @@
...
+        ...infos,
+        race_other: formData.raceOther,
+      };
+    }
+
+    userService
+      .updatePatient(patientId, infos)
+      .then((response) => {
+        if (formData.race) currentPatient.hasRaceAnswer = true;
+        if (formData.takesAnyBloodPressureMedications) currentPatient.hasBloodPressureAnswer = true;
```
MikeD-ZOE, line 180: Ah, so we need each field too to update the currentPatient. (again, future refactor/next steps)

src/core/user/UserService.ts
the-kenneth: Cant this just be a patient.takes_any_blood_pressure_medication != null then?

src/features/multi-profile/SelectProfileScreen.tsx
```diff
@@ -71,7 +71,12 @@ export default class SelectProfileScreen extends Component<RenderProps, State> {
   async startAssessment(patientId: string) {
     const userService = new UserService();
     const currentPatient = await userService.getCurrentPatient(patientId);
-    this.props.navigation.navigate('StartAssessment', { currentPatient });
+
+    if (currentPatient.hasRaceAnswer && currentPatient.hasBloodPressureAnswer) {
+      this.props.navigation.navigate('StartAssessment', { currentPatient });
+    } else {
+      this.props.navigation.navigate('ProfileBackDate', { currentPatient });
+    }
```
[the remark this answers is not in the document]
the-kenneth: Shall we leave it here until the navigation is refactored? Otherwise we'll have to duplicate code in the LevelOfIsolation screen and the healthworker screens too? Wdyt?

assets/lang/en.json
```diff
@@ -109,6 +108,9 @@
     "worked-other-facility": "Other health care facility"
   },
 
+  "back-date-profile-title": "Some additional questions about you",
+  "update-profile": "Update profile",
```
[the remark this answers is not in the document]
abubuwe: I will ask now but might create a channel solely for translations (with Sweden and any other potential countries). How does that sound?

src/core/user/UserService.ts
MikeD-ZOE: That route, you'd probably want to check undefined too, so:
`patient.takes_any_blood_pressure_medication != null || patient.takes_any_blood_pressure_medication != undefined`

assets/lang/en.json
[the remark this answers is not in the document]
MikeD-ZOE: +1

src/features/multi-profile/SelectProfileScreen.tsx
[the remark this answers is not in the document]
MikeD-ZOE: The LevelOfIsolation and Healthworker screens are assessment screens, so should appear after StartAssessment, and when the various pre-requisites (patient info completed, backfilled, etc).

As it stands, Sweden doesn't have multiple profiles, so this logic needs to be duplicated on the WelcomeRepeat screen -- that's why I initially moved this navigation logic to Start Assessment (back when we had multiple WelcomeRepeat screens).

src/core/user/UserService.ts
the-kenneth: != null also covers undefined, so Ill go with that. Makes it more readable

PR #201, opened 2020-04-29
Notch build versions for 2nd build of 0.12

MikeD-ZOE:
# Description

Had to do a second build of the app yesterday, so iOS build picked the correct push notification key from expo. This is the app.json used for the second build.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [X] iOS - Physical device

## Screenshots

### DIFF
--- a/app.json
+++ b/app.json
@@ -26,11 +26,11 @@
     "ios": {
       "supportsTablet": true,
       "bundleIdentifier": "com.joinzoe.covid-zoe",
-      "buildNumber": "0.0.28"
+      "buildNumber": "0.0.29"
     },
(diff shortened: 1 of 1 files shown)


MikeD-ZOE:
> I approved but I'm not clear why we need that for an OTA...

It wasn't for OTA. The first build of the app created new Push Notification keys on the iOS build, and that was uploaded to the App store. When we resolved the issue (in that I shouldn't use my expo login to do the build), I needed to increment iOS buildNumber to do the second iOS build, otherwise the App Store would reject the upload because it already has an app-build with that number.

PR #202, opened 2020-04-29
[Splash] Show an error dialog when startup info fails to fetch

### LINKED ISSUE #31: Handle lack of network connection on splash screen
As the app loads up, the` SplashScreen.tsx` makes an API call. If these user doesn't have an internet connection this will fail and the user just sees a blank screen. 

We don't currently support offline mode - but we do need a warning message.

theblixguy:
# Description

Add an error dialog when we fail to fetch startup info. This should also fix #31 (although it seems we don't show a blank screen anymore). Let me know if you want me to change the wording of the message!

Question: Calling `bootstrapAsync` again shows an error message on the simulator (it seems to be related to changes to the navigator this function makes), but this won't be shown on a prod build. Nevertheless, I was thinking whether we should wait before making changes to navigator, etc?

This is my first time with React Native so apologies in advance for any mistakes!

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

### DIFF
--- a/app.json
+++ b/app.json
@@ -9,7 +9,7 @@
       "ios",
       "android"
     ],
-    "version": "0.12",
+    "version": "0.13",
     "orientation": "portrait",
     "icon": "./assets/icon.png",
     "splash": {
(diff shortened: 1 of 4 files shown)


theblixguy:
@selvach Done!

abubuwe:
Please include the following translations in the sv-SE.json
```
"splash-error": {
    "title": "Fel",
    "message": "Det uppstod ett fel vid hämtningen av startinformation",
    "primary-action-title": "Försök igen",
    "secondary-action-title": "Avbryt"
  },
```

theblixguy:
@abubuwe Done!

PR #203, opened 2020-04-29
Assessment copy changes

MikeD-ZOE:
# Description

Update 5 questions with refined text.

- [ ] Android - Emulator
- [X] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Screenshot 2020-04-29 at 14 12 27


src/core/user/UserService.ts
```diff
@@ -25,7 +25,7 @@ import {
 } from './dto/UserAPIContracts';
 import { camelizeKeys } from './utils';
 
-const ASSESSMENT_VERSION = '1.3.0'; // TODO: Wire this to something automatic.
+const ASSESSMENT_VERSION = '1.3.1'; // TODO: Wire this to something automatic.
```
abubuwe, line 28: should we bump here? this will come out with a series of new questions, so we can just bump to 1.4 when releasing?

MikeD-ZOE: Doing a low-risk OTA this evening, this turns out to be quite precient.

PR #208, opened 2020-04-29
feature(Login screen): Button CTA disabled if fields are empty

emma3333:
# Description

Disable button CTA on Login screen #207

Added function to disable button CTA if input fields are empty.

- [ ] Android - Emulator
- [ ] iOS - Emulator

## Screenshots

Screenshot 2020-04-29 at 16 43 27

Updating other screens with the disabled button logic where appropriate


src/features/login/LoginScreen.tsx
```diff
@@ -41,6 +41,11 @@ export class LoginScreen extends Component<PropsType, StateType> {
     componentDidMount() {
     }
 
+    checkFieldsFilled() {
+        if (this.state.hasPassValidationError || this.state.hasUserValidationError) return false
```
JoshElgar, line 45: This could be `return !(this.state.hasPassValidationError || this.state.hasUserValidationError);`, looks good, something we definitely need.

PR #221, opened 2020-05-03
Date picker poc

selvach:
This is a PoC of a better date picker experience for selecting the date of your COVID Test than the existing implements. 

The current solution uses a native iOS date picker that is deprecated and has pretty poor UX.

This library creates a unified experience for picking 1 date (when a user is certain of their TestDate) and a date range when a user is uncertain of their test dates. 

One major psychological advantage of this approach, is that if a user is struggling to remember a date - displaying a month view with the days of the week (Mon ,Tue...) is a good cognitive aid.

Screenshot 2020-05-03 at 22 19 15
Screenshot 2020-05-03 at 22 19 29


abubuwe:
Thanks for looking at this @selvach! Looks great!

JoshElgar:
Great!

PR #225, opened 2020-05-04
Reset push notification tokens

MikeD-ZOE:
# Description

Refactor of Push Notification code into a more reusable PushNotificationService.
And extend it so that once every 7 days, on the WelcomeRepeat screen, the app sends the pushToken to our backend, to ensure it is active.

* `WelcomeRepeatScreen` and `OptionalInfoScreen` just call `refreshPushToken` and `initPushToken` methods on `PushNotificationService`
* `PushNotificationService` depends on `IPushTokenRemoteClient` (which `UserService` currently implements), and `IStorageService` (which `LocalStorageService` implements by wrapping `AsyncStorage`)
* date handling is refactored into `src/core/utils/datetime.tsx` utility methods.
* platform sniffing is done using `src/core/utils/platform.ts`, a future refactoring is removing the same logic from `src/components/Screen.tsx`

- [X] Android - Emulator
- [ ] iOS - Emulator
- [X] iOS - Physical device

## Screenshots

n/a

## Checklist


src/features/register/WelcomeRepeatScreen.tsx
```diff
@@ -38,23 +29,44 @@ export class WelcomeRepeatScreen extends Component<PropsType, WelcomeRepeatScree
...
+  async updateUserCount() {
     const userService = new UserService();
     const userCount = await userService.getUserCount();
     this.setState({ userCount });
+  }
+
+  async updatePushToken() {
+    const pushTokenService = new PushNotificationService();
+    const aWeekAgo = moment().subtract(7, 'days');
+    let shouldUpdate = false;
```
selvach, line 46: Suggestion: Refactor shouldUpdate() into a function that returns a boolean?

src/features/register/OptionalInfoScreen.tsx
```diff
@@ -44,34 +43,34 @@ export class OptionalInfoScreen extends Component<PropsType, State> {
...
-      const pushToken = await PushNotificationService.getPushToken(false);
+      const pushService = new PushNotificationService();
+      const pushToken = await pushService.createPushToken();
       if (pushToken) {
         try {
-          await userService.savePushToken(pushToken);
-          AsyncStorageService.setPushToken(pushToken);
+          const userService = new UserService();
+          await userService.updatePushToken(pushToken);
+          await pushService.savePushToken(pushToken);
```
julienlavigne, line 54: 🔧 Start both promises and then await then to make the calls asynchronous and get a few ms back :-)

src/features/register/OptionalInfoScreen.tsx
```diff
@@ -44,34 +43,34 @@ export class OptionalInfoScreen extends Component<PropsType, State> {
...
     }
+  }
 
-    const hasFormData = formData.phone?.trim() || formData.name?.trim();
+  private async handleSaveOptionalInfos(formData: OptionalInfoData) {
+    const patientId = this.props.route.params.patientId;
+    const userService = new UserService();
+    const currentPatient = await userService.getCurrentPatient(patientId);
 
+    this.sendPushToken();
```
julienlavigne, line 67: 🔧 Should probably be awaited as it uses AsyncStorage internally

src/features/register/WelcomeRepeatScreen.tsx
```diff
@@ -38,23 +29,44 @@ export class WelcomeRepeatScreen extends Component<PropsType, WelcomeRepeatScree
...
     Navigator.resetNavigation((this.props.navigation as unknown) as NavigationType);
+  }
+
+  async updateUserCount() {
     const userService = new UserService();
     const userCount = await userService.getUserCount();
     this.setState({ userCount });
+  }
+
+  async updatePushToken() {
```
julienlavigne, line 43: 🔧 This should be extracted from the screen. It is fully decoupled from it

src/features/register/WelcomeRepeatScreen.tsx
```diff
@@ -38,23 +29,44 @@ export class WelcomeRepeatScreen extends Component<PropsType, WelcomeRepeatScree
...
         try {
-          await userService.savePushToken(pushToken);
-          AsyncStorageService.setPushToken(pushToken);
+          const userService = new UserService();
+          await userService.updatePushToken(pushToken);
+          await pushTokenService.savePushToken(pushToken);
         } catch (error) {
-          // Ignore failure.
+          // Ignore failure
         }
```
julienlavigne, line 66: 🔧 This code also exists in OptionalInfoScreen, it should be factorized

src/features/register/WelcomeRepeatScreen.tsx
MikeD-ZOE: Or do away with it altogether....

src/features/register/OptionalInfoScreen.tsx
MikeD-ZOE: This approach has a side-effect -- saving to local state assumes the sending the token to the server succeeds. The next refactoring step (eradicating the almost duplicate code on the WelcomeScreen and WelcomeRepeat screen), can lead us into the following scenario:

New user signs up, token is created, send asynchronously to both the server and local storage. The server request fails, the local storage succeeds. The app refreshes the push tokens every week. So for the first week, at least, the user won't be receiving push notifications, because the server doesn't have an active push notification token for them.

With the current both awaiting, the token gets saved to the local storage only when it's happy the server request has completed successfully. I feel that optimising for that is a better choice here.

src/features/register/WelcomeRepeatScreen.tsx
MikeD-ZOE: 👍

src/features/register/WelcomeRepeatScreen.tsx
MikeD-ZOE: 👍

src/features/register/OptionalInfoScreen.tsx
MikeD-ZOE: done

MikeD-ZOE:
Going to break this into multiple PR requests, so it can be reviewed more effectively.

PR #226, opened 2020-05-04
Automated deployments 

selvach:
# Description

Triggers deployment when merging to master
Adds release channels based on major app version (Context variable can be deleted).
Adds slack approvals for master deployment
Fixes the context on stage branch to be the same

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

* Will add hotfix / release branches in a separate PR if this works


.circleci/config.yml
```diff
@@ -100,7 +113,7 @@ workflows:
           filters: *stage_filter
       - deploy:
           filters: *stage_filter
-          context: stage
+          context: react-covid-tracker-stage
```
fmontesino, line 116: changing the context name implies a change in CircleCI, just to make sure we don't lose sight of it

.circleci/config.yml
```diff
@@ -42,16 +43,28 @@ commands:
...
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+              echo 'export RELEASE_CHANNEL=`cat app.json | jq -r '.expo.version' | cut -d. -f1`-prod' >> $BASH_ENV
+            elif [ "${CIRCLE_BRANCH}" == "stage" ]; then
+              echo 'export RELEASE_CHANNEL=`cat app.json | jq -r '.expo.version' | cut -d. -f1`-stage' >> $BASH_ENV
+            elif [ "${CIRCLE_BRANCH}" == "develop" ]; then
+              echo 'export RELEASE_CHANNEL=`cat app.json | jq -r '.expo.version' | cut -d. -f1`-dev' >> $BASH_ENV
+            else
+              echo 'export RELEASE_CHANNEL=$CIRCLE_BRANCH' >> $BASH_ENV
+            fi
```
fmontesino, line 57: This could be easily automated if we had a Context variable in CircleCI, such as $RELEASE_CHANNEL, with the proper values for each Context.
Then you can replace it with:
```
echo 'export RELEASE_CHANNEL=`cat app.json | jq -r '.expo.version' | cut -d. -f1 $RELEASE_CHANNEL >> $BASH_ENV
```

.circleci/config.yml
```diff
@@ -42,16 +43,28 @@ commands:
   login_to_expo:
     steps:
       - run: npx expo-cli login -u $EXPO_USERNAME -p $EXPO_PASSWORD
+      - run:
+          name: Set expo release channel
```
fmontesino, line 47: Would it make more sense to have this inside the "publish to expo" job? Or is it a step purely related immediately after login

.circleci/config.yml
selvach: Yes! We didn't even have the context `stage` setup in CircleCI.  So this was just cleaning up the naming convention. 

@fmontesino - Do you know how to copy / clone a CircleCI context? Or get the ENV variables out of CircleCI? I tried ssh into the CircleCI box and did printenv - but only the project level env variables are available. The context level env variables are not available.

.circleci/config.yml
selvach: So creating contexts appear to be a pain! And we want to have a new release channel when we incremement the major app version

For example: The master branch uses context: react-covid-tracker-prod. But the release channel will be 1.x-prod or 2.x-prod....

.circleci/config.yml
selvach: As the Release Channel is also used deploy_android and deploy_ios, I thought it better to lift it out. Perhaps it should be a separate command

fmontesino: So the steps for the Deploy jobs are:
- build
- publish_to_expo
- deploy_ios
- deploy_android

`publish_to_expo` also includes login to expo as a separate command:
```
steps:
      - login_to_expo
      - run: echo "API_URL=$API_URL" > .env
```
I think it'd be nice to have a command in there to set the release channel as well, IMO not as part of the login command

.circleci/config.yml
fmontesino: But that's what you do with `jq` right? You parse the app.json file, search for the `expo.version` number and then append `-prod` or `-stage` or `-dev`. 

So the version number is dynamic with `jq` and the suffixes are constants for a Context.

.circleci/config.yml
fmontesino: I am afraid there's no copy/clone feature at the moment. However, the idea behind the Context is that you should only put in there those variables with different values for different Contexts. Anything "global and constant" shouldn't be in a Context, as you would have repetition...
So even if you could clone the Context, you'd have to setup the same vars

.circleci/config.yml
```diff
@@ -73,6 +70,8 @@ commands:
           path: ~/covid-react/fastlane/app.apk
   deploy_ios:
     steps:
+      - set_release_channel
+      - login_to_expo
```
fmontesino, line 74: I think you don't need this or the android one. The `deploy` job does everything for you:
```
deploy:
    executor: node
    working_directory: ~/covid-react
    steps:
      - build
      - publish_to_expo.  // <-- This runs login and set release channel
      - deploy_ios
      - deploy_android
```

selvach: Yeah... I was thinking about when people want to independently run deploy_android / deploy_ios. But that's not currently a use case.

PR #237, opened 2020-05-06
App rename

the-kenneth:
Rename text in app and also map pictures


abubuwe:
can you bump version numbers please?

src/components/CovidRating.tsx
```diff
@@ -15,9 +15,9 @@ type State = {
   showTakeToStore: boolean;
 };
 
-const USiOSLink = `https://apps.apple.com/us/app/covid-symptom-tracker/id1503529611`;
-const UKiOSLink = `https://apps.apple.com/gb/app/covid-symptom-tracker/id1503529611`;
-const SEiOSLink = `https://apps.apple.com/se/app/covid-symptom-tracker/id1503529611`;
+const USiOSLink = `https://apps.apple.com/us/app/covid-symptom-study/id1503529611`;
+const UKiOSLink = `https://apps.apple.com/gb/app/covid-symptom-study/id1503529611`;
+const SEiOSLink = `https://apps.apple.com/se/app/covid-symptom-study/id1503529611`;
```
MikeD-ZOE, line 20: This particular change worries me. Do we have any previous experience and knowledge that the store URL will change -- or is that slug a placebo?

the-kenneth: Yes, its a placebo, you can put any text there. as long as the ID is correct

PR #238, opened 2020-05-06
Update multiselect logic

abubuwe:
# Description

When selecting from the multi-select options, selecting None or Prefer Not to say, should de-select the other options.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Unable to demonstrate via a ss

No

### DIFF
--- a/src/features/patient/fields/HormoneTreatmentQuestion.tsx
+++ b/src/features/patient/fields/HormoneTreatmentQuestion.tsx
@@ -39,6 +39,8 @@ const treatmentValues = {
   PREFER_NOT_TO_SAY: 'ht_pfnts',
 };
 
+const singleOptions = [treatmentValues.NONE, treatmentValues.PREFER_NOT_TO_SAY];
+
 type TreatmentCheckBoxData = {
   label: string;
   value: string;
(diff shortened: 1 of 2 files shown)


src/features/patient/fields/RaceEthnicityQuestion.tsx
```diff
@@ -56,8 +56,11 @@ const createRaceCheckboxes = (data: RaceCheckBoxData[], props: FormikProps<RaceE
         value={props.values.race.includes(checkBoxData.value)}
         onChange={(checked: boolean) => {
           let raceArray = props.values.race;
-          if (checked) {
+          if (checkBoxData.value == 'prefer_not_to_say') {
```
selvach, line 59: Please used === and !==

src/features/patient/fields/HormoneTreatmentQuestion.tsx
```diff
@@ -70,8 +70,12 @@ const createTreatmentCheckboxes = (data: TreatmentCheckBoxData[], props: FormikP
         value={props.values.hormoneTreatment.includes(checkBoxData.value)}
         onChange={(checked: boolean) => {
           let treatmentArray = props.values.hormoneTreatment;
-          if (checked) {
+          if ([treatmentValues.NONE, treatmentValues.PREFER_NOT_TO_SAY].includes(checkBoxData.value)) {
+            treatmentArray = [checkBoxData.value];
+          } else if (checked) {
             treatmentArray.push(checkBoxData.value);
+            treatmentArray = treatmentArray.filter((val) => val !== treatmentValues.NONE);
+            treatmentArray = treatmentArray.filter((val) => val !== treatmentValues.PREFER_NOT_TO_SAY);
```
MikeD-ZOE, line 78: You should be able to refactor that into one `filter`. And probably using the same `includes` approach as above. So it makes sense to declare a constant outside this class of `SingleOptions = [treatmentValues.NONE, treatmentValues.PREFER_NOT_TO_SAY];`

`filter()` can then be `filter((val) => !SingleOptions.includes(val));` => "filter returns all items that are not included in the list of SingleOptions"

abubuwe: Thanks for the tip.

PR #251, opened 2020-05-07
Fix Swedish translations

selvach:
# Description

Due to the initialisation of i18n, we need to move the items drop down definition. 

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist


selvach:
Yup - you're right Josh. The have to be inside the class, not the render to get the correct i18n.locale

PR #254, opened 2020-05-07
Remove us optional data screen for normal consent

the-kenneth:
# Description

We want to remove the optional data screen for normal US consent users. This is somewhat of a work around as I am modify the county config on the go, but I think its the simplest way, happy for other suggestions.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

### DIFF
--- a/src/features/Navigation.ts
+++ b/src/features/Navigation.ts
@@ -136,7 +136,12 @@ const ScreenFlow: any = {
     const { patientId } = routeParams;
     const config = navigator.getConfig();
 
-    if (config.enablePersonalInformation) {
+    let askPersonalInfo = config.enablePersonalInformation;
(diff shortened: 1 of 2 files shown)


selvach:
Ugrh.. I understand how this would work and do exactly what the feature requirements as for. But I don't think anyone will be able to intuit why in a few weeks. 

Could we explore a fix that explicitly checks if isUS and isNursesStudy = True and encapsulate the business logic in a function call. 

It's a little sad this can't be done solely through countryConfig.

the-kenneth:
I agree with you Somesh, and am happy to do a little more exploring to see if I can get something cleaner. If not i should comment it at least

the-kenneth:
Ive found a somewhat cleaner approach, this should be better

selvach:
Thanks @the-kenneth !

PR #263, opened 2020-05-11
Fix to allow date to be changed for covid test

the-kenneth:
# Description

It is currently not possible to switch between a single date and a range of dates. This fixes that by clearing the date(s) of the opposite selection when switching between the question of the knowDateOfTest question.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device


selvach:
Screenshot 2020-05-12 at 13 41 29

PR #272, opened 2020-05-13
Refactor StartAssessment into navigator

the-kenneth:
# Description

Removes the start assessment screen and puts the logic into the navigator. Still many screens are not using the navigator/now using both however this is a start.

@MikeD-ZOE I wasn't sure if i should be using setNavigation/resetNavigation, it works as is atm but I couldn't quite figure if I should be using those.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

Next steps would be to get others screens using the navigator and move more logic into the navigator by using goToNextScreen rather than individual functions.


selvach:
LGTM - worked through the logic in your commits. But have not tested.

src/features/Navigation.ts
```diff
@@ -92,6 +92,51 @@ class Navigator {
...
+      return CommonActions.reset({
+        index: 0,
+        routes: [...newStack],
+      });
+    });
+
+    this.startAssessment(currentPatient, null);
+  }
+
+  startAssessment(currentPatient: PatientStateType, assessmentId: string | null) {
```
selvach, line 111: I wonder if style wise, it's nicer to make assessmentId an optional parameter? i.e. `assessmentID?: string`
Does it then pass undefined when you use it below in `this.navigation.navigate()`?

Take it or leave it :)

the-kenneth:
Had to merge manually to have forced push up my merge fixes

PR #273, opened 2020-05-14
Add Amplitude Analytics

selvach:
# Description

Add's Amplitude Analytics - initial PR tracks the Share This App & Donate.

N.B. It's not possible to see on Android whether they actually shared the message, only that they opened the share module. iOS does provide some information on the callback, but it's not particularly interesting. 

AMPLITUDE_KEY is passed in via CircleCI 
App is functional and non-blocking if the API key isn't present or working. 
All events are "fire and forget".

Amplitude support have disabled the user fields we don't want to capture at the project level. (This is supported programatically, but only after upgrading to expo sdk 0.37)

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist


selvach:
@MikeD-ZOE Can we talk about this line tomorrow?
https://github.com/zoe/covid-tracker-react-native/blob/20cf2a3ae7f097fe8c41206525a66d0581ef4083/src/core/user/dto/UserAPIContracts.ts#L44

I added is_tester to the definition. But the whole object is stringified and saved to asyncStorage.

MikeD-ZOE:
> I added is_tester to the definition. But the whole object is stringified and saved to asyncStorage.

As long as the document is JSON.stringify, that should be fine. -- JSON functions can handle boolean values fine.

PR #274, opened 2020-05-14
Fix race backfill logic

the-kenneth:
# Description

Fix for logic and rename variable to make it less confusing. Unfortunately we are using the same backend variable for 2 slightly different things so it cant be perfectly simple but I hope this is better.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

### DIFF
--- a/src/core/user/UserService.ts
+++ b/src/core/user/UserService.ts
@@ -163,7 +163,7 @@ export default class UserService extends ApiClientBase {
     patient: PatientInfosRequest
   ): Promise<PatientStateType> {
     // Calculate the flags based on patient info
-    const hasRaceAnswer = patient.race.length > 0;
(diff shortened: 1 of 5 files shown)


src/core/user/UserService.ts
```diff
@@ -224,7 +224,7 @@ export default class UserService extends ApiClientBase {
       isFemale,
       isPeriodCapable,
       isHealthWorker,
-      hasRaceAnswer,
+      hasRaceEthnicityAnswer: hasRaceAnswer,
```
abubuwe, line 227: why not just fully rename the variable above.

the-kenneth: Yeh done, just a javascript gotcha

PR #275, opened 2020-05-14
Stop multiple submit taps for adding tests

the-kenneth:
# Description

Due to the relatively low numbers of people who have duplicates, and the amount of duplicates per person (most are 3 or less duplicates). This seems to be an issue of frustrated button tapping when the servers are unresponsive.

This fix stops the sending of tests multiple times if the button is tapped more than once.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device


abubuwe:
I don't see how you'd resubmit with this logic.
There's no way to set submitting back to false.
You might also find some benefit in using the built-in formikProp for this - further explanation here https://jaredpalmer.com/formik/docs/guides/form-submission

PR #276, opened 2020-05-14
Add dual US locale

the-kenneth:
# Description

For the US checks if the user has es or en in their locale and uses for app locale.

Also adds all es translations

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device


the-kenneth:
@MikeD-ZOE What are the rules regarding i18n strings in different countries? The tests have failed when merging in from develop and I think we might need to switch around a lot of the strings we create to different files.

MikeD-ZOE:
> @MikeD-ZOE What are the rules regarding i18n strings in different countries? The tests have failed when merging in from develop and I think we might need to switch around a lot of the strings we create to different files.

Since the Study is UK only, I think the translation strings for that feature should be in `en-GB.json`.

We do need a discussion to create rules about what strings need to go where, if just for consistency. Ideally, we should be checking that translations are available for features that country needs.

We have translation strings that are used based on the country, but we don't currently reflect that in our translation strings structure.

Considering the i18n importing of JSON is straight requires, it could be possible to break up the translation json into features, and each locale can aggregate together the translation string it needs.

We need a sensible default to test against, though. I'm not convinced the i18n tests we have are right, because we aren't comparing apples to apples. The correct set of i18n strings to test against should be directly tied to the features and functionality of that country. The assumption that `en` is an accurate representation of what in the app needs translating is flimsy, and makes translators do more work than is required.

abubuwe:
@MikeD-ZOE @the-kenneth Do we need an es-US.json as well, and then the i18n testing could compare both the base es, en, sv files for consistency, and also compare files for the same countries, e.g US?

the-kenneth:
I think adding es_US and converting tests would be done in a separate PR if it were to happen

PR #277, opened 2020-05-15
Improve error handling

MikeD-ZOE:
# Description

Network/Server errors come in two forms: ones that can be retried, and ones that can't or shouldn't. Errors because the User is offline, or the server is busy (429 errors) are those that can be retried. 404 errors can't be retried, the resource isn't there, retrying isn't going to change things.

When the user is presented with a retry button, we're wrapping the handler of that in a 5 second pause, but indicating that the app is retrying the request. That way, we are doing a basic throttling of requests, giving the server at least a little time to scale up to handle the current volume of requets. We can evolve that delay in future, e.g. increasing the delay after each failed request.

### Offline/Server down error-handling:

* SplashScreen - instead of a plain blue screen, there's now a Covid logo, and underneath that a status message. If requests for startup info fail (most usually because the user is offline or our server is busy), we'll display a status message underneath the logo, and after a brief pause a retry button. This eliminates the blue screen of death, and offers a recoverable way out of network failures. I've done a refactor of the processing logic to break it down into modular/componentised steps, some of this logic should be moved into their respective services as a next step.
* SelectProfileScreen - instead of an empty screen when the request for the patient list fails, we show a message with the error, and if the request is retryable, a retry button to trigger off a delayed retry of the request.
* The Bob problem: Bob appears as a profile name when the request to get the Patient document from the server fails, it currently falls back to an empty patient state, which has the default name of Bob. Now we catch when the request for the patient document fails, we present a modal dialogue displaying an error, and if the request is retryable, they have a Retry button which triggers off a delayed retry. If the request is not retryable, an OK button closes the modal, and the user remains on the same page -- here they might have the ability to go a different way.

There are five routes that can trigger the appearance of Bob:
* SelectProfileScreen -- request for that selected profile fails
* Create profile, when ConsentForOthersScreen is completed, and the request for that newly created profile fails
* OptionalInfoScreen -- in navigating to the next page retrieving the patient details of the main profile user fails
* WelcomeRepeatScreen -- when multiple profiles is turned off (e.g. Sweden), then clicking on Reports Symptoms to navigate to the first page of assessment, the request to retrieve the main profile patient details fails.

The cleanest set of changes for showing how error-handling was added to the screen is best shown on the WelcomeRepeatScreen and ConsentForOthersScreen.

The fifth way is through Create/RegisterScreen in countries that have the OptionalInfo screen turned off (this will shortly be the US). This is outstanding, and I prefer to do this at the same time as improving the error handling on the Create Account and Login pages.

In each of these screens, the logic to catch an error, and to retry the request feels like it can be refactored into something reusable, but I haven't found the right way yet. It feels like screens extending from a custom subclass of React Component is the right way. Though I suspect someone will say React Hooks :-D

### Unresolved / partially complete issues

`src/core/ApiServiceErrors` is a collection of Exceptions I use to wrap API errors. The main feature is to expose properties such as isRetryable, so we can offer the option to try again to the user. This does duplicate the library of Exceptions in `src/core/Exception`. Perhaps the next step is to refactor the new file to just have API-error-wrapping Exceptions (which is what `handleServiceError` does).

I had started a feature to track whether the user is Offline or our server was unavailable. This is currently in `src/core/offline/OfflineService.tsx`, with React component `src/components/ErrorMessageBar.tsx` and `src/components/OfflineNotice.tsx`. I'm stuck in how to best surface that, not just to display, but how to update from UserService, or how to manage the feature after it's determined the use is offline or our server is unavailable. I've left it in, and removed the original usage from `CovidApp.tsx`.

### Refactoring UserServices

`src/core/Services` is a singleton that exposes instantiated services, so we can get an instance of UserService and other services instead of instantiating it each time. That takes us one step closer to writing unit tests on services, because we have space to inject mock objects without complicating the React Native layer. 

I had a go in UserService of defining interfaces for the various Services in there that could be broken out into separate independent services. The idea is we can move out the interface to a new file, and all the named methods in UserService, then refactor that to be self contained. Before that, we need to refactor the calls to the Api into generic method calls via an API client -- UserService needs to stop being an API Client, and use an API client given to it.

- [ ] Android - Emulator
- [X] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

SplashScreen:

Screenshot 2020-05-15 at 12 08 37

SelectProfileScreen:

Screenshot 2020-05-15 at 12 11 07

Catching Bob on SelectProfileScreen:

Screenshot 2020-05-15 at 12 11 34

## Checklist

* Swedish translations, @abubuwe 
* Further work on improving error handling on Login and Create account screens.
* Continuing to refactor User Services into separate self-reliant services
* Refactoring Splashpage setup logic into appropriate service instances.

### DIFF
--- a/src/core/ApiServiceErrors.ts
+++ b/src/core/ApiServiceErrors.ts
@@ -13,6 +13,20 @@ type ReceivedError = {
   response?: AxiosResponse;
 };
 
+export type ApiErrorState = {
+  isApiError: boolean;
+  error: AppException | null;
+  status: string;
+  onRetry?: () => void;
+}
+
+export const initialErrorState = {
+  isApiError: false,
+  error: null,
+  status: '',
+}
+
+
 export class AppException extends Error {
   friendlyI18n: string | null;
   isRetryable = false;
--- a/src/core/ApiServiceErrors.ts
+++ b/src/core/ApiServiceErrors.ts
@@ -3,7 +3,9 @@ import { AxiosResponse } from 'axios';
 const NETWORK_ERROR = 'Network Error';
 
 const STATUS_NOT_FOUND = 404;
+const STATUS_REQUEST_ERROR = 400;
 const STATUS_SERVER_BUSY = 429;
+const STATUS_SERVER_ERROR = 500;
 
 type ReceivedError = {
   message: string;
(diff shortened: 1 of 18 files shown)


src/core/ApiServiceErrors.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (error.isAxiosError && error.response) {
+    switch (error.response.status) {
+      case STATUS_NOT_FOUND:
+        throw new ApiException(error.message, error.response.status, 'errors.resource-not-found');
+      case STATUS_SERVER_BUSY:
+        throw new RetryableApiException(error.message, error.response.status, 'errors.server-is-busy');
+      case STATUS_SERVER_ERROR:
+        throw new RetryableApiException(error.message, error.response.status, 'errors.server-error');
+      default:
+        console.log('[ERROR] Unhandled status:', error.response.status, ':', error.message);
```
selvach, line 65: Do you intend to keep this in?

src/core/offline/OfflineService.ts
```diff
@@ -0,0 +1,93 @@
+import appConfig from '../../../appConfig';
+import axios from 'axios';
+import moment from 'moment';
+
+// const ONLINE_URL = 'https://www.google.com/robots.txt';
+const ONLINE_URL = 'https://joinzoe.com/icons/icon-48x48.png';
```
selvach, line 6: I would change this - Hannah is going to deco joinzoe.com and create brand new Company page.

selvach:
Amazing stuff - this makes a lot of sense and thank you for the detailed description. This feels like a good place to draw the line. I agree with you on potentially refactoring some of the handling of errors on the 3 main screens - but later on!

src/core/ApiServiceErrors.ts
MikeD-ZOE: Yes, for the time being, so I can track any other useful status codes we need to track.

src/core/offline/OfflineService.ts
MikeD-ZOE: Good to know. The "background isOffline" check isn't currently wired up (well, it was, then I removed it), so it's not being used yet.

the-kenneth:
Reviewing...

abubuwe:
Requested Swedish translations FYI.

MikeD-ZOE:
merge conflicts resolved.

PR #286, opened 2020-05-17
changes to README.md

mnecchi:
# Description

I tried to setup the project in order to study it and eventually contribute to it, but as soon as I started the app I got an error for a missing `AMPLITUDE_KEY` variable.
I've changed the README file adding the step to add a test value to the `.env` file.

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,13 +75,19 @@ npm install expo-cli --g
    echo "API_URL=http://123.456.7.890:3000" > .env
    ```
 
-4. Create an empty `./google-services.json` file in the root of the application.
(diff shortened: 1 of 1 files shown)


selvach:
Thank you very much. I'll see if I can make that parameter optional.

mnecchi:
@selvach If you manage to find a way let me know please.
I've tried a couple of times but I couldn't find a way.
My initial idea was to do something like `import * as env from 'react-native-dotenv` and then check `env.WHATEVER` to see if it's undefined or not, but it doesn't work.

PR #287, opened 2020-05-17
Fix/mock server

mnecchi:
# Description

This is a basic refactor of the mock server to give it the capability of storing and retrieving data from json files.

After setting up the project I tried to run a simple flow in the app to start understanding how it works and what I can work on.
Soon I discovered that many api calls where returning 404 and the others where returning just static data.
Basically it was not possible to complete an assessment, add new tests and create new profiles (which were stuck to the three static ones), etc.
I understand this is a development environment but I think having a behaviour so far from the real one can cause a lot of issues and bugs.

So I decided to extend the mock server.
This is very far from perfect but at least now you can test the app more thoroughly.

At the moment only the endpoints for patients, assessments and covid tests are implemented with the json storage.

- [x] Android - Emulator
- [x] iOS - Emulator
- [x] iOS - Physical device

## Checklist
- [x] I have updated mockServer

Extended the json storage to other endpoints


mnecchi-akkroo:
> looks good to me.

@MikeD-ZOE Hey Mike, I've realised I've made a mistake after your review: the server was supposed to preserve the state after restart but there was a small bug.
I've also changed `fs.exists` to `fs.access`...didn't know exists is deprecated...

MikeD-ZOE:
> > looks good to me.
> 
> @MikeD-ZOE Hey Mike, I've realised I've made a mistake after your review: the server was supposed to preserve the state after restart but there was a small bug.
> I've also changed `fs.exists` to `fs.access`...didn't know exists is deprecated...

No worries. Thank you for your contributions, and for catching these.

This reminds me too. although `fs.exists` is deprecated, `fs.existsSync` isn't. The bootstrap function is fairly critical for the server to work, so that it goes async opens up possible race conditions. Two that I can imagine from the above code:

mnecchi-akkroo:
> This reminds me too. although `fs.exists` is deprecated, `fs.existsSync` isn't. The bootstrap function is fairly critical for the server to work, so that it goes async opens up possible race conditions. Two that I can imagine from the above code:

Yep. I thought about that, but, for the specific use case, I opted for prompt startup against fail safe.
I explain: it's true that the bootstrap is critical, but what are the chances we start writing/reading to the mock server straight away?
Usually we manually start the mock server, then we start expo, then the app, then the app loads the js bundler, then...by the time we start interacting with the server the bootstrap has already done its job. Also because it should do it only the first time or when a new "table" is added.

Anyway if you prefer a sync bootstrap (which shouldn't be that slow anyway) I can change it.

selvach:
I think this is great work and worth merging in, so that the rest of dev team can benefit from it. If we feel like we need to switch the sync bootstrap, we can do that at a later!

PR #291, opened 2020-05-18
Module aliases (with bulk change of imports)

mnecchi:
# Description

This is a proposed change to enhance the dev experience when importing modules.
The aim is to avoid the [relative path hell](https://goenning.net/2017/07/21/how-to-avoid-relative-path-hell-javascript-typescript-projects/).

For example in `src/features/login/LoginScreen.tsx` instead of
```
import UserService, { isUSCountry } from '../../core/user/UserService';
```
you can use
```
import UserService, { isUSCountry } from '@covid/core/user/UserService';
```
The module aliases introduce are:
- `@covid` (-> src/)
- `@assets` (-> assets/)
- `@theme` (-> theme/)

The PR has a huge number of files because I did a bulk replace using regex to update all the imports.
The only files worth reviewing carefully are the ones in the first commit [6b8a2e2](https://github.com/zoe/covid-tracker-react-native/commit/6b8a2e2cd41167a8a00296a30cad3379a860f2c0), which are:

- `tsconfig.json`
- `package.json`
- `src/package.json`
- `assets/package.json`
- `theme/package.json`

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Checklist


selvach:
Thank you for this @mnecchi! I love how this is going to make the dev experience much better. I'll look to the merge this during a "quiet" period in the next few days. 

We're paying down a lot of our technical debt accrued as we launched quickly - so this is a really welcome enhancement!

mnecchi:
> Thank you for this @mnecchi! I love how this is going to make the dev experience much better. 

I'm glad you think it's useful @selvach !

> We're paying down a lot of our technical debt accrued as we launched quickly - so this is a really welcome enhancement!

I totally understand. Covid couldn't wait for proper specs! 🤣 
There quite a few things that can be improved.
The thing that really "worries" me is this sort of custom-made state management persisted on AsyncStorage. I'm sure I don't know the rationale behind but I can't really see why you didn't go with Redux or other already made solutions.
It's true that Redux  has a bit of boiler plate code (not so much you had gone with hooks to be honest) but it gives you well documented and tested solution with a lot of tools already available.

src/core/user/UserService.ts
```diff
@@ -2,9 +2,9 @@ import { AxiosResponse } from 'axios';
 import * as Localization from 'expo-localization';
 import moment from 'moment';
 
-import { isAndroid } from '../utils/platform';
-import i18n from '../../locale/i18n';
-import { AvatarName } from '../../utils/avatar';
+import { isAndroid } from '@covid/utils/platform';
```
abubuwe, line 5:
```suggestion
import { isAndroid } from '@covid/core/utils/platform';
```

mnecchi: Yep. Sorry, I tried to be smart but I guess I failed miserably! 🤣 
I used a couple of regex to do the changes through out the code. I checked them one by one before applying but I guess I've missed some. My bad.

selvach:
@mnecchi Perhaps you could split this into sepearte PRs? One that adds the json configurations in your first commit, which we can merge immediately.

Then we can take small passes at changing the imports across the repo?

mnecchi:
> @mnecchi Perhaps you could split this into sepearte PRs? One that adds the json configurations in your first commit, which we can merge immediately.

Yes, it makes total sense! 👍 
I'll do it as soon as I have some spare time.

src/features/register/OptionalInfoScreen.tsx
```diff
@@ -6,16 +6,17 @@ import React, { Component } from 'react';
...
+import { AsyncStorageService } from '@covid/core/AsyncStorageService';
+import { PiiRequest } from '@covid/core/user/dto/UserAPIContracts';
+import i18n from '@covid/locale/i18n';
 import Navigator from '../Navigation';
 import { ScreenParamList } from '../ScreenParamList';
-import { ApiErrorState, initialErrorState } from '../../core/ApiServiceErrors';
-import { userService, offlineService, pushNotificationService } from '../../Services';
-import { LoadingModal } from '../../components/Loading';
+import { ApiErrorState, initialErrorState } from '@covid/core/ApiServiceErrors';
+import { userService, offlineService, PushNotificationService } from '../../Services';
```
abubuwe, line 18:
```suggestion
import { userService, offlineService, pushNotificationService } from '../../Services';
```

mnecchi:
@selvach @abubuwe Here is the other PR with config only: https://github.com/zoe/covid-tracker-react-native/pull/305

selvach:
@mnecchi Would you be able to resolve conflicts and we can merge this in? Otherwise, if you drop me the regexes, I can take a look!

mnecchi:
@selvach Conflicts resolved

PR #292, opened 2020-05-18
Create API client

MikeD-ZOE:
# Description

Starts an API client that wraps APIClient base, with generic HTTP methods so we can separate out the services from the backend API they persist data. This implements just enough to support PushNotificationService.

Every service that needs to communicate with the Backend API will talk to a service specific API client, which in turn talks to this API Client, which wraps Axios. Because this API client wraps the ApiClientBase it inherits the userToken management, so we can do authenticated requests without each service having to deal with it itself.

This is necessary step to break apart User Service in a meaningful way,

- [ ] Android - Emulator
- [X] iOS - Emulator
- [X] iOS - Physical device

## Screenshots

## Checklist

### DIFF
--- a/src/core/api/ApiClient.ts
+++ b/src/core/api/ApiClient.ts
@@ -1,0 +1,21 @@
+import { ApiClientBase } from '../user/ApiClientBase';
+import { handleServiceError } from '../ApiServiceErrors';
+
+export interface IApiClient {
+  post<T>(path: string, object: T): Promise<T>;
+}
+
+export default class ApiClient extends ApiClientBase implements IApiClient {
+  protected client = ApiClientBase.client;
+
+  async post<T>(path: string, payload: T): Promise<T> {
+    try {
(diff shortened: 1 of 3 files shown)


abubuwe:
Thanks for also fixing that circular import!

src/core/api/ApiClient.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export default class ApiClient extends ApiClientBase implements IApiClient {
+  protected client = ApiClientBase.client;
+
+  async post<T>(path: string, payload: T): Promise<T> {
+    try {
+      const response = await this.client.post<T>(path, payload);
+      console.log("[RESPONSE]:", response);
+      return response.data;
```
mnecchi: Apart from the `console.log` why not destruct `response`?
The code can be:
```
const { data } = await this.client.post<T>(path, payload);
return data;
```
Not a big change but a bit more readable I think.

PR #293, opened 2020-05-18
Fix keyboard above input fields

the-kenneth:
## Description

Move keyboard avoiding view up into screen object to wrap the scroll view. When it was nested it wasn't working.

Main changes in the Screen file. The other changes are removing the lower level keyboard avoiding view and then reformatting.

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

The Screen component needs a general fix up. We need to not wrap everything in a scroll view and also handle the header navigation better.


selvach:
The diff tool in InteliJ / Webstorm is far better for this kind of thing....

PR #296, opened 2020-05-18
Fix LevelOfIsolation screen appearing weekly.

abubuwe:
# Description

Flip diff on `shouldAskLevelOfIsolation`. It was previously evaluating as negative and never evaluating to true for existing users. 

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device


selvach:
I know it's a one liner. But this is crying out to be turned into a function and a unit test added, please.

selvach:
Done! Spotted a very similar function used in the DaysAgo component.

abubuwe:
Realising this now, but does the explicit conversion to utc() not cause the possible date localisation bug on the profile screen @selvach?

selvach:
@abubuwe Backend always returns UTC. So this code does what it should. 

I think that 1 report by 1 user is somewhat related to timezones. But it might not be.... I'd rather not investigate.

selvach:
@MikeD-ZOE Changes are ready for you to review.

PR #310, opened 2020-05-20
Add share medium to tracking

selvach:
# Description

Adds ShareMedium from the callback, if it's present. 

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device
- [x] Amplitdue 

## Screenshots

## Checklist


src/components/ShareThisApp.tsx
```diff
@@ -21,11 +21,13 @@ export default class ShareThisApp extends Component<Props> {
...
     const message = this.shareMessage + (isAndroid ? ' ' + this.shareUrl : ''); // On Android add link to end of message
     try {
-      await Share.share({
+      const result = await Share.share({
         message,
         url: this.shareUrl, // IOS has separate field for URL
       });
-      Analytics.track(events.SHARE_THIS_APP);
+
+      const sharedOn = result.action === Share.sharedAction && result.activityType ? result.activityType : 'unknown';
```
MikeD-ZOE, line 29: Android won't return a `results.activityType`, that only works on iOS.

I'd suggest that instead of just returning `unknown`, that we check `isAndroid` here, and send back `android`, and leave `unknown` for iOS devices that should have the field populated.

selvach:
@MikeD-ZOE Thanks for the suggestion. Changes ready for review!

PR #313, opened 2020-05-20
Stage Release. 

selvach:


package.json
```diff
@@ -112,10 +112,5 @@
       "node_modules/(?!(jest-)?react-native|react-clone-referenced-element|@react-native-community|expo(nent)?|@expo(nent)?/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|@sentry/.*)"
     ]
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm test"
```
MikeD-ZOE, line 117: Instead of removing the `npm test` entirely can we do something like attaching it to a hook around push? 

I'd like these tests to be automatically running, just in their current slow state, just not on git commit

selvach: They'll run automatically in CI. What value would you get by putting this on a push hook?

MikeD-ZOE: They are the only form of automated testing we have running in a development environment. The value is not forgetting to run them.

mnecchi: > They'll run automatically in CI. What value would you get by putting this on a push hook?

@selvach The value is that you can errors quickly and avoid a useless run for CircleCI.

I agree with @MikeD-ZOE : there's no need to run tests for every commit (sometimes I want to commit something that's broken intentionally) but definitely before pushing the branch.
Also the more tests we add the more it will take to run them, so waiting a couple of minutes for every commit is not the best. 🤣 

In my current job I've implemented a git hook that runs lint, prettier and tests before pushing. If you want I can "move" that hook here.

selvach: Sounds like a plan. We don';t need this PR any more.

mnecchi: This is a nice talk about this topic (and other interesting things as well): https://www.youtube.com/watch?v=ikn_dBSski8

Anyway, I will give it a go later and open a PR when done.

PR #315, opened 2020-05-21
Global mocks

mnecchi:
# Description

### UserService mock

Writing the first couples of tests I've found out that the UserService mock was implemented in all the tests and my hunch is that it will be used in many more tests to come.
For this reason I've decided to extract it and make a "global" mock in `src/core/user/__mocks__`.

In order to use this mock you should just include the following in the test (this is for the `login` method):
```ts
import * as UserService from '@covid/core/user/UserService';

jest.mock('@covid/core/user/UserService');
const { mockedLogin } = UserService as any;
```

### native-base mock

I've created a mock for `native-base`, this is in the `__mocks__` under root and so it will be automatically loaded by jest.
For the moment it only contains the mocked implementation for `Toast.show`.

### fix to module aliases

I've modified slightly the aliases rules in the `tsconfig.json` file: I don't know if it is just my vscode but the old ones didn't seem to work with ts-jest.

Hope this is useful.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

Update the existing tests to use this mocks


selvach:
Thanks for this! Happy to merge @MikeD-ZOE

PR #327, opened 2020-05-22
Update consent text

the-kenneth:
# Description

Update validation study consent text


abubuwe:
Thanks for this @the-kenneth !

PR #329, opened 2020-05-25
Tests and refactor for Welcome Screens

### LINKED ISSUE #84: Create Unit test for Welcome Page
Scenarios for unit test:

Welcome Screen
a) Visit NHS website link
b) Create account button
c) More info link

mnecchi:
# Description

Fix #84 

This PR includes tests and refactor for the Welcome Screen.
I've started with Welcome1Screen, which was, to my surprise, implemented as a Hook.
As you probably know Enzyme shallow doesn't really work well with Hooks. (if you don't know and you are interested there's a long discussion in this [issue](https://github.com/facebook/react/issues/17321)).
There are a couple of workarounds to make Enzyme shallow work with Hooks but they are all extremely hacky.
I usually prefer to ditch Enzyme straight away and use [react-native-testing-library](https://github.com/callstack/react-native-testing-library) (I think it's, in general, easier and clearer than Enzyme anyway)

After installing the library I started writing the tests for Welcome1Screen.
Once the tests were done and all passing I decided to refactor it:
- I moved it into a separate folder to split the logic from the styles
- I slightly modified the code because it was not following the best practices for Hooks. For example, it was using an object with the state, which can create problems and nasty bugs if you do shallow copies of the object when assigning it (not a problem in this specific case but better to get into the habit). Better to have multiple `useState` with "simple" values.
- I also removed inline function declaration and used `useCallback` hook for every callback used.

Then I moved to Welcome2Screen.
First of all I wrote the tests and when done and all the tests were passed I decided to convert the code using hooks, just to align it with Welcome1Screen.
I also removed the code in `componentDidMount` because not used here (I guess it was a copy-paste from the other screen).
The refactor for Welcome2Screen followed the same logic and steps of the other screen.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I have run `npm test`


src/features/register/Welcome1Screen/Welcome1Screen.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const getMapImage = useCallback(() => {
+    if (isGBCountry()) {
+      return gbMap;
+    }
+    if (isSECountry()) {
+      return svMap;
+    }
+    return usMap;
+  }, [isGBCountry, isSECountry, gbMap, svMap, usMap]);
```
MikeD-ZOE, line 40: Considering this callback function doesn't use anything in the current closure, it makes sense to move its implementation out of this component and reduce this to a one-liner:

```
const getMapImage = useCallback(getLocaleMapImage);
```

src/features/register/Welcome1Screen/Welcome1Screen.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const getFlagIcon = useCallback(() => {
+    if (isGBCountry()) {
+      return gbFlag;
+    }
+    if (isSECountry()) {
+      return svFlag;
+    }
+    return usFlag;
+  }, [isGBCountry, isSECountry, gbFlag, svFlag, usFlag]);
```
MikeD-ZOE, line 50: Same here. nothing in the callback implementation relies on the current closure scope.

src/features/register/Welcome2Screen/Welcome2Screen.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const onCreateAccountPress = useCallback(async () => {
+    if (await userService.shouldAskCountryConfirmation()) {
+      setIpModalVisible(true);
+    } else {
+      if (isUSCountry()) {
+        navigation.navigate('BeforeWeStartUS');
+      } else {
+        navigation.navigate('Consent', { viewOnly: false });
+      }
```
MikeD-ZOE, line 36: This navigation logic should be in Navigator (src/Navigation.tsx), probably as part of `gotoNextScreen()`

src/features/register/Welcome2Screen/Welcome2Screen.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const flagIcon = useCallback(() => {
+    if (isGBCountry()) {
+      return gbFlag;
+    }
+    if (isSECountry()) {
+      return svFlag;
+    }
+    return usFlag;
+  }, [isGBCountry, isSECountry, gbFlag, svFlag, usFlag]);
```
MikeD-ZOE, line 48: This is a duplicate of what's on Welcome2Screen. Can the implementation be put in one place and reused/called here?

src/features/register/Welcome1Screen/Welcome1Screen.tsx
mnecchi: If the `getLocaleMapImage` is used elsewhere it makes sense (you should tell me, I don't know the codebase as you yet 😄 ). If not I don't see any improvement in doing that.

src/features/register/Welcome1Screen/Welcome1Screen.tsx
mnecchi: This is different because, I think, the same code is used in `Welcome2Screen`, so it can be shared somewhere.

src/features/register/Welcome2Screen/Welcome2Screen.tsx
mnecchi: I totally agree but this is how it was done before. I don't know if there was a rationale behind.
If you want to have all the navigation in one place (which makes total sense to me) I think also the login screen (and probably many others) must be changed.

src/features/register/Welcome2Screen/Welcome2Screen.tsx
mnecchi: Definitely

src/features/register/Welcome2Screen/Welcome2Screen.tsx
mnecchi: Rethinking about that...I'm not sure...
Having all specific screen navigation rules outside the screen itself probably makes maintaining harder.
Also it makes the `Navigation.tsx` to grow huge making it hard to maintain as well.
For example, I really think the `CovidApp.tsx` is already too big to be maintained easily, I would probably split in many different files if possible.

PR #330, opened 2020-05-25
Node tests for jest expo enzyme preset

mnecchi:
# Description

This PR adds the possibility to run tests for the node platform with the jest-expo-enzyme preset.
The idea is that if you have a library or an helper which is just js/ts code and as nothing to do with native development, it doesn't make much sense to test it twice for iOS and Android platforms.
I think it's better to test it once as a node library/code.
Since we are writing more and more tests and we will execute them for each push with a git hook, we should try to run only meaningful tests and the least we can.
Testing for example `UserService` class for both iOS and Android doesn't make sense because the code is independent from the platforms.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I have run `npm test`

@selvach @MikeD-ZOE I think we should decide a common strategy for tests.
For example: 
- Do we group the tests in a `__tests__` or `tests` folder? Or do we put tests in the same folder as code?
- If we go for `__tests__` folders, where do these folders live? In the same folder of the code tested, in a global folder?
- Do we call our test with `.spec.ts` or `.test.ts`? For example, I've used `.test.ts` because I've seen `src/components/BrandedButton/__tests__/BrandedButton.test.native.tsx`. But now I've noticed that Mike created a `src/locale/compare.spec.ts`. Both are fine let's just decide which one to use.

Then I need to ask about `src/locale/compare.spec.ts`.
Honestly I don't understand the sense of this test, but maybe I'm missing something.
It seems to me that it runs an actual check on the locale configuration.
But this is not the purpose of tests. Tests should check that a piece of code does what it is supposed to do, no more than this.

If you want to run an actual check you should run it locally via a git hook, for example, or add it as a step in a CircleCI job.

@MikeD-ZOE Am I missing something?


mnecchi:
@selvach @MikeD-ZOE What about the "conventions" questions in the potential follow up?

PR #338, opened 2020-05-27
i18n tests

mnecchi:
# Description

I've modified the test for `src/locale/compare.ts` to make it test the functionalities instead of running the actual locales test.
I've added the locales test to a `pre-push` git hook (via Husky) so that the check is run each time we push a branch.
Also added the check to the CircleCI `build_and_test` job.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I have run `npm test`


scripts/i18n-check-locale.ts
```diff
@@ -1,11 +1,10 @@
 import { compareLocale } from '../src/locale/compare';
 
 const args = process.argv;
-const locale: string = args.length > 1 ? args[2] : '';
+const locale: string = args.length > 2 ? args[2] : '';
```
mnecchi, line 4: Are we sure we need to run the script without locale?

selvach:
@MikeD-ZOE I've resolved conflicts, are you happy with this?

PR #346, opened 2020-05-28
Spanish i18n tests + fixes

abubuwe:
Adds some missing ES translations.
This PR also adds the CI test for checking Spanish translations.

As a result, I have removed the en-GB file, and moved the US only strings from the en-US file into the main en in #347


abubuwe:
That was painful!

PR #349, opened 2020-05-28
Add Babel Editor 

selvach:
# Description

This PR uses BabelEditor to manage the json files. No more manually editing the json! 

BabelEditor will ensure that new keys are added to all json files. You can even rename and move (nest) them entire in the UI

Unfortunately this does sort the keys - and we have many that are in the top level. I plan to group and nest this in a series of separate PRs. (Greatly simplified). 

If you guys are down with this, I'll arrange licenses. 

https://www.codeandweb.com/babeledit

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


assets/lang/sv-SE.json
```diff
@@ -1,812 +1,738 @@
...
+    "header-15": "",
+    "header-16": "",
+    "header-2": "",
+    "header-3": "",
+    "header-4": "",
+    "header-5": "",
+    "header-6": "",
+    "header-7": "",
+    "header-8": "",
+    "header-9": "",
```
MikeD-ZOE, line 385: The ordering of translation strings by key (why???) means these translation strings are too difficult to maintain, because they are no longer in a logical order as they are found in the screen these appear on. The grouping together of paragraphs with their respective headings, which gives context to them, has gone out of the window.

I am struggling to understand how this would make sense to a non-technical user, let alone remove technical people from the translations strings loop. This feels broken by design.

assets/lang/sv-SE.json
```diff
@@ -1,812 +1,738 @@
...
+    "loading-data": "",
+    "methodology-body-1": "",
+    "methodology-body-2": "",
+    "more-people": "",
+    "news-feed": "nyhetsflöde",
+    "of-residents": "",
+    "our-methodology-title": "",
+    "out-of": "",
+    "people-with-covid-in": "",
+    "places-since-yesterday": "",
```
MikeD-ZOE, line 562: And this--- empty strings. How does a non-technical user know it's because the translation is missing, or the translation doesn't exist, or it's an error? I can guess, I can also be wrong. And that wasn't the point of putting in automated checking.

This makes a complete mockery of the i18n unit test. What was the point of the i18n unit test if not to protect our codebase from missing translation strings. This puts back the problem the i18n unit test was supposed to help us with - fix missing translation text.

assets/lang/sv-SE.json
selvach: This is something the UI does automatically. 

We have a lot of translations strings that are not nested. I started trying to nest them per screen and per functionality - but stopped, as the PR would become unusable. My plan is to re-nest them in separate PRs. 

All the Translation Managers I looked at did this unfortunately :(

I don't envy this task, to be honest. And it might be a reason to stop this avenue.

PR #351, opened 2020-05-29
New vitamin questions

abubuwe:
# Description

Adds new question on vitamins. 

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

Possible tool-tip style info icon to describe why we're adding this question.


src/features/patient/YourHealthScreen.tsx
```diff
@@ -160,6 +171,7 @@ export default class YourHealthScreen extends Component<HealthProps, State> {
     const userService = new UserService();
     var infos = this.createPatientInfos(formData);
 
+    console.log(infos);
```
selvach, line 174: Needs to be removed...

MikeD-ZOE:
I think this question is in the wrong place. It's not a profile question. It feels like a lifestyle question where the answer changes over time.

Is this a question that should be asked periodically? If it's once off, I suspect the question text needs to be narrowed to "recently". And for the purpose of backfilling the profile, the date and time the user filled in the question should be recorded.

abubuwe:
> I think this question is in the wrong place. It's not a profile question. It feels like a lifestyle question where the answer changes over time.
> 
> Is this a question that should be asked periodically? If it's once off, I suspect the question text needs to be narrowed to "recently". And for the purpose of backfilling the profile, the date and time the user filled in the question should be recorded.

There are currently no plans to ask this periodically. But I agree that based on the current schema, because this question is relative to when it is asked, we should track when it was asked, so I have added a field for this. 

And I agree that checkbox questions like this should be refactored in a seperate PR @selvach - there were a lot of places where it felt like I was tripping over myself.

PR #356, opened 2020-05-29
[Refactor] Nav clean up

the-kenneth:
# Description

As discussed, I've looked at how to clean the navigation up and I specifically looked at starting with the assessment workflow.

Ive taken inspiration from the iOS coordinator pattern in order to decouple screens from their navigation logic. This approach will also allow the screen to be decoupled from handling its own state as well, but for now Ive just started with navigation.

I have an 'AssessmentCoordinator' class that is similar to our 'Navigation' class. However it also holds the props that are passed to each screen in the 'AssessmentData' variable, this can be expanded out in the future to hold onto the whole state.

Initially I wanted to pass the whole coordinator object through to each screen, but as react throws warnings about passing non-serialisable objects (it does work though...) I opted to have the coordinator as a singleton for now.

There are some clean ups/small fixes i still need to do, but just wanted to put this out now that the first pass done. 

- [x] iOS - Emulator

I think this can be used to manage state quite easily as a layer above whatever we want to use, for now though it can manage it on its own by converting AssessmentData to AssessmentState and holding all the user inputted data.

I don't really like having a singleton because an init() function needs to be called each time at the start of the flow in order to put the current assessment data in.


src/features/assessment/AssessmentCoordinator.ts
```diff
@@ -0,0 +1,171 @@
...
+  assessmentId?: string;
+  currentPatient: PatientStateType;
+};
+
+class AssessmentCoordinator {
+  navigation: NavigationType;
+  userService: UserService = new UserService();
+  assessmentData: AssessmentData;
+
+  ScreenFlow: any = {
```
selvach, line 28: YAASS! This is what I was trying to explain in the Tech Call, but couldn't find the right words to explain :) 💯

selvach:
[1] @the-kenneth Do you think we could write a short ADR to document your coordinator pattern approach.  I've seen projects where ADRs are checked into the repo, but we also have a few in Notion already from the ZOE project. 

The second paragraph in your PR description concisely sums it up!

(https://github.com/joelparkerhenderson/architecture_decision_record#suggestions-for-writing-good-adrs) 

[2] Can we write some tests around the AssessmentCoordinator? This way we can show "what great looks like".

src/features/assessment/AssessmentCoordinator.ts
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  startAssessment = () => {
+    const { currentPatient } = this.assessmentData;
+
+    const userService = new UserService();
+    const features = userService.getConfig();
+
+    if (currentPatient.hasCompletedPatientDetails) {
+      const mustBackfillProfile =
```
selvach, line 95: Perhaps refactor this out into a smaller pure function, that can easily have a unit test.

src/features/assessment/AssessmentCoordinator.ts
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      const nextPage = this.getPatientDetailsScreenName(currentPatient);
+      this.gotoScreen(nextPage);
+    }
+  };
+
+  gotoEndAssessment = async () => {
+    const config = this.getConfig();
+    const shouldShowReportForOthers =
+      config.enableMultiplePatients &&
```
selvach, line 122: Refactor as well?

src/features/assessment/AssessmentCoordinator.ts
MikeD-ZOE: Yes please, so we can move it -- in good time -- to a PatientService. :-)

src/features/Navigation.ts
```diff
@@ -20,6 +20,45 @@ class Navigator {
   navigation: NavigationType;
   userService: UserService;
 
+  ScreenFlow: any = {
```
MikeD-ZOE, line 23: There's a `ScreenFlow` here in Navigation, and there's a `ScreenFlow` in NavigationCoordinator. Is one meant to replace the other, or are they both going to remain?

If one will prevail over the other, what's currently blocking that? (I'm not expected that to be done right now, just want a feel on what's outstanding/problematic/issue)

selvach:
Can we also benefit from Nested Navigators here?

src/features/Navigation.ts
the-kenneth: My intention was that each 'flow' would have its own coordinator with similar `ScreenFlow` objects. 

We have 3 main flows in the app: assessment, patient and home (the main flow from the start of the app that starts the other two).

We have yet to refactor the navigation file but I was intending it to become the `HomeCoordinator` (naming up for debate).

the-kenneth:
> Can we also benefit from Nested Navigators here?

I tried this for the sake of organisation but found it added unnecessary complexity. And then after I checked the React Navigation website (I know.. should rtfm first!) and it states what I kind of found out:

> Think of nesting navigators as a way to achieve the UI you want rather than a way to organize your code. If you want to create separate group of screens for organization, keep them in separate objects/arrays rather than separate navigators.

So we **should** organise our screens better by splitting them up, but nesting is not needed as our UI doesnt call for it.

src/features/Navigation.ts
selvach: > Think of nesting navigators as a way to achieve the UI you want rather than a way to organize your code. If you want to create separate group of screens for organization, keep them in separate objects/arrays rather than separate navigators.
> 
> So we **should** organise our screens better by splitting them up, but nesting is not needed as our UI doesnt call for it.

Amazing - thanks for the insight. Might be worth slinging that into an ADR. @julienlavigne was asking about this earlier.

PR #369, opened 2020-06-02
Send locale on register

the-kenneth:
# Description

Sends the 2 char locale string on register. In draft waiting on backend


selvach:
Given this is just on the registration endpoint, only new Spanish users will get push notifications in Spanish. Existing ones will receive them in English.... and I'm kind of okay with that! 

Existing users will have been using the rest of the App in English all this time. There aren't that many of them. We save doing an extra patch / put for millions of users on the welcome screen. 

The focus in the US / MGH is acquiring new Spanish speaking minorities.

the-kenneth:
> The server will only accept es, en, sv and null. Any other values will result in 400. Are we happy that we won't block users from signing up (e.g. if the are using the UK app with French locale).

We limit the locale in the app only to use those values so its fine.

PR #370, opened 2020-06-02
[REFACTOR] AssessmentService and use it in Assessment pages

MikeD-ZOE:
# Description

AssessmentService has two methods:
* `saveAssessment` for saving the assessment on each page
* `completeAssessment` to mark an assessment as completed

Currently `saveAssessment` does what the current UserService does -- POSTs or PATCHes the assessment on each screen. And `completeAssessment` calls `saveAssessment` if there's an assessment document.

This is to prepare for the next step, adding redux to the AssessmentService, so that `saveAssessment` stores the assessment updates in state, and `completeAssessment` will then POST the aggregated assessment document to the backend API.

Each screen's handle method(s) have changed in the following thematic ways:
* They are async, and use the anonymous method way of binding `this`
* Promised converted to await - cleaning up the code quite nicely.
* `assessmentService.saveAssessment()` called instead of `UserService.addAssessment()` or `UserService.updateAssessment()` -- `assessmentId` being `null` is how we know it's a new assessment, currently
* calls to `Navigator.gotoEndAssessment()` are immediately preceded with `AssessmentService.completeAssessment()` -- that's the hook we need for an "End of Assessment" state.
* On screens with forked paths that include ending the assessment, I've kept the `updateAssessment` helper method, and added an `isComplete` flag to indicate whether the callee branch is an end of assessment path or not.

- [ ] Android - Emulator
- [X] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


MikeD-ZOE:
Moving out of draft.

src/features/assessment/HealthWorkerExposureScreen.tsx
```diff
@@ -53,28 +53,20 @@ export default class HealthWorkerExposureScreen extends Component<HealthWorkerEx
...
-        .catch((err) => this.setState({ errorMessage: i18n.t('something-went-wrong') }));
+  handleUpdate = async (formData: HealthWorkerExposureData) => {
+    try {
+      const { assessmentId } = AssessmentCoordinator.assessmentData;
+      var assessment = this.createAssessment(formData);
+
+      const response = await assessmentService.saveAssessment(assessmentId!, assessment);
+      if (!assessmentId) {
+        AssessmentCoordinator.assessmentData.assessmentId = response.id;
+      }
```
julienlavigne, line 64: 💭 I'm worried about this pattern of setting the state from the screen. This code is not obvious to me. Once redux is in place, AssessmentCoordinator should be refactored rapidly to subscribe to the store instead.

src/features/assessment/HowYouFeelScreen.tsx
```diff
@@ -34,25 +36,34 @@ export default class HowYouFeelScreen extends Component<HowYouFeelProps, State>
     AssessmentCoordinator.resetNavigation(props.navigation);
   }
 
-  handleFeelNormal = () => {
-    this.updateAssessment('healthy')
-      .then(() => AssessmentCoordinator.goToNextHowYouFeelScreen(true))
-      .catch(() => this.setState({ errorMessage: i18n.t('something-went-wrong') }));
+  handleFeelNormal = async () => {
+    if (await this.updateAssessment('healthy', ASSESSMENT_COMPLETED)) {
```
julienlavigne, line 40: 💭 I'm guessing that ASSESSMENT_COMPLETED is introduced purely for code clarity, is that right?
I don't like the fact that it's a file-level const and thus in scope for the whole file if only needed in one place.
Alternatively, you could:
- declare the const locally
- have 2 different functions (updateAssessment and completeAssesment). Those have rather different behaviour anyway since downstream one post to the server and can reasonably fail with the other just update local state and shouldn't

src/features/assessment/HowYouFeelScreen.tsx
```diff
@@ -34,25 +36,34 @@ export default class HowYouFeelScreen extends Component<HowYouFeelProps, State>
...
   };
 
-  private updateAssessment(status: string) {
-    const assessmentId = AssessmentCoordinator.assessmentData.assessmentId;
-    const userService = new UserService();
-    const promise = userService.updateAssessment(assessmentId!, {
-      health_status: status,
-    });
-    return promise;
+  private async updateAssessment(status: string, isComplete: boolean = false) {
```
julienlavigne, line 51: 💭 This is actually more a tryUpdateAssessment since a possible return value is "false". This sounds really C-ish and could benefit from modelling the error case more explicitely. I guess the obvious pattern is to have a promise instead and chaining with .then(x => ...)

src/features/assessment/LevelOfIsolationScreen.tsx
```diff
@@ -102,23 +103,20 @@ export default class LevelOfIsolationScreen extends Component<LocationProps, Sta
...
-      .catch(() => this.setState({ errorMessage: i18n.t('something-went-wrong') }));
+  async handleUpdate(formData: LevelOfIsolationData) {
+    try {
+      const { currentPatient, assessmentId } = AssessmentCoordinator.assessmentData;
+      var assessment = this.createAssessment(formData);
+
+      const response = await assessmentService.saveAssessment(assessmentId!, assessment);
+      if (!assessmentId) {
+        AssessmentCoordinator.assessmentData.assessmentId = response.id;
+      }
```
julienlavigne, line 114: 💭 I've seen that code before in HealthWorkExposureScreen

the-kenneth:
Reviewing still, please dont merge yet

src/core/assessment/AssessmentService.ts
```diff
@@ -0,0 +1,70 @@
...
+      version: ASSESSMENT_VERSION,
+    };
+    return this.apiClient.post<AssessmentInfosRequest, AssessmentResponse>(API_ASSESSMENTS, assessment);
+  }
+
+  updateAssessment(assessmentId: string, assessment: AssessmentInfosRequest): Promise<any> {
+    const assessmentUrl = `/assessments/${assessmentId}/`;
+    return this.apiClient.patch<AssessmentInfosRequest, AssessmentResponse>(assessmentUrl, assessment);
+  }
+}
```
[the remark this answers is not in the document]
MikeD-ZOE: The `apiClient` is shared across the various services (PushNotificationService remote client uses it too, and I have the ContentService in progress that already uses it). It's convenient because it retains the user auth token seamlessly.

src/features/assessment/HealthWorkerExposureScreen.tsx
MikeD-ZOE: This code will disappear once the app is posting one request at the end of the assessment, since we won't have an assessmentId to pass around until the end.

src/features/assessment/HowYouFeelScreen.tsx
MikeD-ZOE: Yeah, `true` doesn't tell me much what the second parameter does, and why it's that value. So I use a constant to create something more explanatory. We can do it locally, no worries.

src/features/assessment/LevelOfIsolationScreen.tsx
MikeD-ZOE: Yes, it's on all assessment screens that could be "first assessment screen" to post to the server and then get back an assessmentId that the following pages need to patch the right document.

src/core/assessment/AssessmentService.ts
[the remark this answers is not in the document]
julienlavigne: I was talking about the AssessmentApiClient, sorry for the confusion

the-kenneth:
Is there a potential issue with multiple asyncs being fired off if pressing the button more than once?

MikeD-ZOE:
> Is there a potential issue with multiple asyncs being fired off if pressing the button more than once?

Potential, no. It's already happening. On the HealthcareExposureScreen, for example. We do need a nice repeatable pattern to stop that. I keep thinking that subclassing React.component is the way, but... ugh.

Hopefully, when the "Do one post at the end of the assessment" is in place, the app should be noticeably more responsive, which will minimise the time before someone septuple-taps the next question button.

MikeD-ZOE:
> Also are we using arrow functions or not? This is something that should be standardized I think as this inconsistent in the codebase but also within the PR. Unless there is a logic I didn't perceive?

I ran into an issue yesterday early on that adding `async` to a method that was explicitly binded to `this` in the constructor broke. The error message sounded looked like it couldn't determine `this.updateAssessement()`, so I switched to using the arrow function and it worked fine. I assumed it was because `async` and `bind()` didn't work together well. And I followed the same pattern across all the Assessment flow screens.

But trying to recreate the problem now, it seems to work... So I don't know what the missing broken piece was.

src/features/assessment/HowYouFeelScreen.tsx
MikeD-ZOE: You are right, there's a C-smell there. I didn't want callers to this function handle any thrown errors, because they'd do the same thing, so I did it in the `updateAssessment()`, and just returning a boolean to indicate it was done or not.

I should just re-`throw` the error instead. That's the proper exception handling approach.

I'm trying to discipline myself to just use async/await, and not mix them with promise/then/catch. I got called out on a technical coding exercise for mixing them together -- which was a fair criticism at the time.

Updating this HowYouFeelScreen and  WhereYouAreScreen with this revised approach.

PR #373, opened 2020-06-03
Fix showing errors before submission

selvach:
# Description

- Fixes errors being shown on Your Work and About You screen before submission. 
- Adds check on formik props.touched and formik submitCount > 0

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 8 - 2020-06-03 at 13 21 29

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/components/GenericTextField.tsx
+++ b/src/components/GenericTextField.tsx
@@ -36,7 +36,7 @@ export const GenericTextField = (props: GenericTextFieldProps) => {
         />
       </Item>
 
-      {showError && !!formikProps.errors[name] && (
+      {showError && !!formikProps.touched[name] && !!formikProps.errors[name] && (
         <ValidationError
           // @ts-ignore - need to solve type for ValidationError error prop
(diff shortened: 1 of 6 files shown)


src/components/GenericTextField.tsx
```diff
@@ -36,7 +36,7 @@ export const GenericTextField = (props: GenericTextFieldProps) => {
         />
       </Item>
 
-      {showError && !!formikProps.errors[name] && (
+      {showError && !!formikProps.touched[name] && !!formikProps.errors[name] && (
```
abubuwe, line 39: Does this solve the error of it being shown before submission?
Seems to only solve for issues once edited. Is that intended?

selvach: If you start typing a postcode, then delete your entry and move on, it will show an error (just on the postcode field) stating this is required. 

Which I think is an improvement.

PR #374, opened 2020-06-03
Cleanup Share Module

selvach:
# Description

- Remove unused prop on Share Module. 
- Fix imports

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Screenshot 2020-06-03 at 19 27 51

## Checklist

- [ ] I have run `npm test`


abubuwe:
Are we really losing this functionality? Not sure I agree that it's not going to be useful going forwards unless you think otherwise?

selvach:
I'm 95% confident it's not coming back. It was temporary whilst we had Donations live.

abubuwe:
Simple enough change to revert if we need to so I'll run with it :)

PR #376, opened 2020-06-03
Adds New UK Thank You Screen

selvach:
# Description

[1] Adds an Experimental  UK Thank You Screen
- Add External Callout modules / links to content
- Includes tracking in Amplitude 
- Updates Share This App image
- Adds UTM codes on ExternalCallouts so we can track in GA. 

[2] Adds Experiments (A/B Testing) to allow 4 different variants to be shown on the Thank You Page

[3] Add a COVID Ad controlled by a variant of the eligibility endpoint that returns true even if you've previously declined

N.B. - The COVID Ad is likely to be temporary (1-2 weeks) and then will be removed. 

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots
Screenshot 2020-06-04 at 12 27 09

Screenshot 2020-06-04 at 12 27 13

## A/B Testing 
- You will get one of these 4 variants assigned returned to you (and added to you Amplitutde userProperties) when you call `startExperiment()`

image

Screenshot 2020-06-04 at 14 29 33

## Checklist

- [ ] I have run `npm test`

- CTA is still based on the old logic, but will need to test if user has multiple profiles. 
- Webinar weblink needs to be update after the webinar!


src/features/ThankYouUKScreen.tsx
```diff
@@ -0,0 +1,180 @@
...
+                <View style={styles.socialIconContainer}>
+                  <Image source={studyInvite} style={styles.socialIcon} />
+                </View>
+              )}
+
+              <View style={styles.content}>
+                <RegularText style={styles.signOff}>{i18n.t('thank-you-uk.sign-off')}</RegularText>
+              </View>
+
+              <BrandedButton
```
abubuwe, line 106: The appearance of this should be conditional on having more than 1 profile no?

selvach: Not addressed in this PR: 
- "CTA is still based on the old logic, but will need to test if user has multiple profiles."

abubuwe: Ah thanks for clarifiying  - but I think it should still be other way round. Design shows "return to home screen" should be a button and "report for others" should be clickable text.

selvach: Updated!
Screenshot 2020-06-04 at 14 09 50

src/core/Analytics.ts
```diff
@@ -69,9 +75,42 @@ export function identify(additionalProps?: AdditionalUserProperties): void {
...
+
+  const variant = getVariant(profile.username, totalVariants);
+  const payload: { [index: string]: string } = {};
+  payload[experimentName] = variant;
+
+  initialize();
+  Amplitude.setUserProperties(payload);
+  return variant;
+}
+
```
MikeD-ZOE, line 109: Is the Analytics code the right place for a multi-variant testing setup code?

src/features/Navigation.ts
```diff
@@ -148,6 +160,11 @@ class Navigator {
     this.navigation.reset(navStack);
   }
 
+  hasMultipleProfiles(): boolean {
+    // TODD: Determine if user has unreported profiles
+    return false;
```
MikeD-ZOE, line 165: The TODO text conflicts with the method name. What is this function supposed to do?

src/features/ThankYouUKScreen.tsx
```diff
@@ -0,0 +1,204 @@
+import { Text } from 'native-base';
+
+import { blog001, incidence001, studyInvite, surveyInvite, webinar001 } from '@assets';
```
MikeD-ZOE, line 3: This feels like we're missing an abstraction layer that manages an array of content.

src/features/ThankYouUKScreen.tsx
```diff
@@ -0,0 +1,204 @@
...
+      await Navigator.gotoNextScreen(this.props.route.name, { patientId });
+    } catch (error) {
+      // Pass
+    }
+  };
+
+  render() {
+    const showIncidenceCallout = this.state.variant === 'variant_1' || this.state.variant === 'variant_4';
+    const showWebinarCallout = this.state.variant === 'variant_2' || this.state.variant === 'variant_4';
+    const showBlogCallout = this.state.variant === 'variant_3' || this.state.variant === 'variant_4';
```
MikeD-ZOE, line 65: Ah okay, we're building up a multi-variant system... We can refactor this in due course.

src/core/Analytics.ts
selvach: Good point! I've split these out.

src/features/Navigation.ts
selvach: I've killed this code. It was too hard to implement the logic so it was descoped. 

Basically there needs to be a CTA on the Screen:

If user only has 1 profile -> goto WelcomeScreen (CTA Display "Back to home screen"
if user has >1 profile that has not reported today -> goto Select Profile (CTA display: "Report for others"

So we need a helper function at the business layer, not necessarily navigation - as we need to change the CTA display text.

src/features/ThankYouUKScreen.tsx
selvach: Agreed. I struggled to name the files. Lets see what we learn from the experiment!

src/core/Analytics.ts
selvach: Regarding the hash - I believe it should be nicely distributed as it's hashing a UUID!

PR #377, opened 2020-06-04
Add assessment state

MikeD-ZOE:
# Description

* Moved `AssessmentApiClient` into it's own file
* Created `ReduxAssessmentState` that wraps redux dispatchers with a cleaner interface
* Added an `initAssessment` which at the start clears down the assessment state. I couldn't just delete the assessment after it's been posted to the server, because currently the user can navigate back to the assessment and make changes. I handled that by recording the assessment id in the state, so if the id was present, it would PATCH to the server if a user resubmitted the assessment. Meaning it was better to clear the assessment at the start of the assessment. Doing it at the end creates two issues when navigation back into a completed assessment: Duplicate assessments posted, or HTTP 500 errors because of missing patient id field.
* I had a bit of a futz to get the correct type in the reducer actions/events. Not painful, but possibly just the starting point of understanding the kinds of issues we might have without type-safe actions. Not worried yet, I wouldn't mention it, apart from wanting to document possible issues that type-safe actions may solve for us in a possible future/alternate universe.
* I don't like the ASSESSMENT_VERSION hiding out in `src/core/assessment/AssessmentApiClient.ts` -- So next steps are to move it somewhere global/top-level, along with PATIENT_VERSION. FIXED IN ANOTHER PR.
* Spotted an issue with IsolationScreen sending back `NaN` values to the server when not entered.
* The `initAssessment` is done as part of the startAssessment logic in the `AssessmentCoordinator`. To allow the existing unit test to work properly, I injected the `assessmentService` into the static class on the `init()` call -- the same way `userService` is injected.

- [ ] Android - Emulator
- [X] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [X] I have run `npm test`


src/core/assessment/AssessmentApiClient.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  addAssessment(assessment: AssessmentInfosRequest): Promise<AssessmentResponse> {
+    assessment = {
+      ...assessment,
+      version: ASSESSMENT_VERSION,
+    };
+    return this.apiClient.post<AssessmentInfosRequest, AssessmentResponse>(API_ASSESSMENTS, assessment);
+  }
+
+  updateAssessment(assessmentId: string, assessment: AssessmentInfosRequest): Promise<any> {
```
the-kenneth, line 27: Should return an explicit type: Promise<AssessmentResponse>

MikeD-ZOE: Ah, good spot. Fixing.

MikeD-ZOE:
> Now that the assessment service is embedded in the coordinator, I think it would be better for the save and update calls to go through the coordinator so that there is one point of entry to everything to do with the assessment. Wdyt?

That sounds like a good step forward. Though, wouldn't that require the transformed form data to be sent to the AssessmentCoordinator -- or how would it access it?

Does the AssessmentCoordinator recognise when the assessment is complete -- e.g. based on data from the various endpoints?

the-kenneth:
> > Now that the assessment service is embedded in the coordinator, I think it would be better for the save and update calls to go through the coordinator so that there is one point of entry to everything to do with the assessment. Wdyt?
> 
> That sounds like a good step forward. Though, wouldn't that require the transformed form data to be sent to the AssessmentCoordinator -- or how would it access it?
> 
> Does the AssessmentCoordinator recognise when the assessment is complete -- e.g. based on data from the various endpoints?

For now I was just thinking to wrap the save and complete calls to the service in the coordinator, so the service isnt access directly from the screens.

MikeD-ZOE:
> For now I was just thinking to wrap the save and complete calls to the service in the coordinator, so the service isnt access directly from the screens.

Lets do this in a follow up PR?

PR #381, opened 2020-06-04
Add new info page to validation study

the-kenneth:
# Description

Adds page in between intro and consent. Basic styling done.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

More complete styling can be done.


src/features/Navigation.ts
```diff
@@ -149,7 +149,7 @@ class Navigator {
   }
 
   async profileSelected(mainProfile: boolean, currentPatient: PatientStateType) {
-    if (isGBCountry() && mainProfile && (await userService.shouldAskForValidationStudy())) {
```
selvach, line 152: We should still apply this check!

PR #384, opened 2020-06-05
Enforce import linting + run lint

abubuwe:
Enforces `import/order` as an error rather than a warning.
Also lints the repo.

To be re-run once #378 is merged.


assets/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/order */
 // Please keep these in alphabetical order ^_^
```
MikeD-ZOE, line 2: Ah the irony. Turning off alphabetical ordering of imports in a file where the first line is a comment to please preserve the alphabetic order. :-D

PR #387, opened 2020-06-05
Wire ThankYouAd Properly

selvach:
# Description

Wire ThankYouAd to the correct Study Info screen / interstitial screen. 
Passes in an empty PatientState object (null patientID) which means that after Consenting to the Study you land back on the Welcome Screen. 

Somewhat hacky - but this ad goes away in 1-2 weeks. 

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


src/components/InviteToStudy.tsx
```diff
@@ -4,16 +4,17 @@ import Navigator from '@covid/features/Navigation';
...
   render() {
     return (
       <TouchableWithoutFeedback
         onPress={() => {
           Analytics.track(events.CLICK_STUDY_AD_CALLOUT, { placement: this.props.placement });
-          Navigator.gotoScreen('ValidationStudyConsent', {
-            viewOnly: false,
-          });
+          const patientID = '';
+          const currentPatient = getInitialPatientState(patientID);
```
MikeD-ZOE, line 16: This will return a patient state where the patient name is Bob (e.g. default patient values). Is that what we want?

If you want the patient details for the current patient, you'll want to be calling `UserService.getCurrentPatient()`

selvach: Thanks for the rubber duck earlier! 🦆

PR #392, opened 2020-06-05
Inverse hayfever symptoms question

the-kenneth:
# Description

The question was asked inversed compared to all others, so we retranslated and switched it to match.

### DIFF
--- a/src/features/assessment/DescribeSymptomsScreen.tsx
+++ b/src/features/assessment/DescribeSymptomsScreen.tsx
@@ -46,7 +46,7 @@ const initialFormValues = {
   hasDelirium: 'no',
   hasEyeSoreness: 'no',
   isSkippingMeals: 'no',
-  typicalHayfever: 'yes',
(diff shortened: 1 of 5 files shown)


src/features/assessment/DescribeSymptomsScreen.tsx
```diff
@@ -175,7 +175,7 @@ export default class DescribeSymptomsScreen extends Component<SymptomProps, Stat
       red_welts_on_face_or_lips: formData.hasRedWeltsOnFace === 'yes',
       blisters_on_feet: formData.hasBlistersOnFeet === 'yes',
       eye_soreness: formData.hasEyeSoreness === 'yes',
-      typical_hayfever: formData.typicalHayfever === 'yes',
+      typical_hayfever: formData.hasUnusualHayfever === 'yes',
```
abubuwe, line 178: this should be no + only populated if user has hayfever

PR #397, opened 2020-06-07
Add Storybook

selvach:
# Description

This PR adds Storybook and Storybook Server to the repo: https://storybook.js.org/
Storybook is a useful tool to encourage Component Based Design. 

Approach:
 - I've chosen to put a flag to switch from App mode to Storybook mode in `App.tsx`. The other approach is to have an Admin Menu somewhere in the app that takes you to a "developer only" screen, which displays the Storybook. 

- I've collected all stories in one directory under `src/storybook/stories`. It might be better to have each component in `src/components` in its own directory and co-locate the tests, snapshots and stories for that component in a single directory. 

- I've included basic decorators as an example. These a good ways to write functionality that can be injected across all stories (e.g. CenteredView) or add at the story level, which applies to all variations. 

## In Future PRs I'll play with some of the `addons`: 
 - knobs - allowing you to supply different props to the component
 - actions - allows event handlers
 - notes - allows documenting components

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Screenshot 2020-06-07 at 09 21 33

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -1,2 +1,5 @@
 import CovidApp from '@covid/CovidApp';
-export default CovidApp;
+import StorybookUIRoot from '@covid/storybook';
+
+const ENABLE_STORYBOOK = true;
+export default ENABLE_STORYBOOK ? StorybookUIRoot : CovidApp;
--- a/App.tsx
+++ b/App.tsx
@@ -1,5 +1,5 @@
 import CovidApp from '@covid/CovidApp';
 import StorybookUIRoot from '@covid/storybook';
 
-const ENABLE_STORYBOOK = true;
+const ENABLE_STORYBOOK = false;
 export default ENABLE_STORYBOOK ? StorybookUIRoot : CovidApp;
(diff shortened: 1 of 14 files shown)


App.tsx
```diff
@@ -1,2 +1,5 @@
 import CovidApp from '@covid/CovidApp';
-export default CovidApp;
+import StorybookUIRoot from '@covid/storybook';
+
+const ENABLE_STORYBOOK = false;
```
abubuwe, line 4: should this be in the .env file

selvach: I did do that initially - but there's an annoying quirk with react-native's dotenv implementation - it breaks if you don't have the value in the env file; whereas most implementations you don't need it. 

This would break CircleCI - and I'm terrified of us accidentally OTA'ing the Storybook!

At least this way, if you check in App.tsx with `ENABLE_STORYBOOK` = true, CircleCI tests would fail and the you can't deploy.

PR #433, opened 2020-06-17
Add US study invite modal

the-kenneth:
# Description

Feature complete on the front but needs test with the backend before merge.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

image


src/components/USStudyInvite.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  componentDidMount() {
+    const { currentPatient } = this.props.assessmentData;
+    if (isUSCountry() && !currentPatient.isReportedByAnother && currentPatient.shouldShowUSStudyInvite) {
+      this.setState({ isModalOpen: true });
+    }
+  }
+
+  handleAgree = () => {
```
abubuwe, line 39: Can we have amplitude tracking of agreements and disagreements please?

src/core/user/UserService.ts
```diff
@@ -293,6 +293,7 @@ export default class UserService extends ApiClientBase
 
     const hasAtopyAnswers = patient.has_hayfever != null;
     const hasHayfever = patient.has_hayfever;
+    const shouldShowUSStudyInvite = patient.contact_additional_studies;
```
abubuwe, line 296: semantics feel off here - this 'shouldShow' would be true when the field = `null`, however if a user has never seen this before, under this logic, they would never see it.

selvach:
@the-kenneth My screen looks a little different:

Screenshot 2020-06-18 at 11 36 24

I've tested locally with the backend in place. 
This logic needs to be revered as Abu says:
`    const shouldShowUSStudyInvite = !patient.contact_additional_studies;`

It returns `null` initially for all users - so we should ask the user then and set it as true or false. 
After that, we never ask them again.

src/core/user/UserService.ts
```diff
@@ -293,6 +293,7 @@ export default class UserService extends ApiClientBase
 
     const hasAtopyAnswers = patient.has_hayfever != null;
     const hasHayfever = patient.has_hayfever;
+    const shouldShowUSStudyInvite = !patient.contact_additional_studies;
```
abubuwe, line 296: This should just be `patient.contact_additional_studies === null` no?- otherwise users who've declined would see it again?

src/core/Analytics.ts
```diff
@@ -22,6 +22,8 @@ export const events = {
   DECLINE_STUDY: 'DECLINE_STUDY',
   CLICK_STUDY_AD_CALLOUT: 'CLICK_STUDY_AD_CALLOUT',
   CLICK_CALLOUT: 'CLICK_CALLOUT',
+  JOIN_STUDY_CONTACT: 'JOIN_STUDY_CONTACT',
```
abubuwe, line 25: Super minor but I think Accept is better than Join here

src/components/USStudyInvite.tsx
```diff
@@ -11,66 +11,55 @@ import { userService } from '@covid/Services';
...
-  componentDidMount() {
-    const { currentPatient } = this.props.assessmentData;
+  useEffect(() => {
+    const { currentPatient } = props.assessmentData;
     if (isUSCountry() && !currentPatient.isReportedByAnother && currentPatient.shouldShowUSStudyInvite) {
-      this.setState({ isModalOpen: true });
+      setModalVisible(true);
     }
-  }
+  });
```
cm-hvg, line 27: I believe you can do this, so `useEffect` only gets triggered on `props. assessmentData` changed.

```
useEffect(() => {
   // Logic
}, [props.assessmentData])
```

PR #451, opened 2020-06-22
Update one of the lifestyle answer to include unsure.

cm-hvg:
# Description

Add Unsure as an option of Lifestyle Questions

Quick front end change to language only. Change "Prefer not to say" → "Unsure / Prefer Not to Say".

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [x] I have run `npm test`

### DIFF
--- a/appConfig.ts
+++ b/appConfig.ts
@@ -6,4 +6,5 @@ export default {
 
   assessmentVersion: '1.5.0',
   patientVersion: '1.5.0',
+  lifestyleVerison: '1.0.1',
 };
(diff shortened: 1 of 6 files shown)


abubuwe:
Is there a lifestyle version field that we could increment? cc @selvach

selvach:
There is - when you post, you can set the `version` field which you can set to a string "1.0.1"

cm-hvg:
I've added `version` to the API call within `AssessmentService.ts`. Let me know is this the correct way of doing it.

selvach:
Could you move the '1.0.1' to a constants file here:
https://github.com/zoe/covid-tracker-react-native/blob/develop/appConfig.ts

cm-hvg:
Yes i can sir.

PR #452, opened 2020-06-23
Fix border styles on dropdowns

selvach:
# Description

Fixes border issues on dropdowns. The corners were caused by incorrectly identifying the top and bottom most option due a type error in the function signature. 

## Before

Untitled (1)

## After
Screenshot 2020-06-23 at 07 15 41

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


src/components/DropdownField.tsx
```diff
@@ -93,12 +93,13 @@ export const DropdownField: React.FC<DropdownFieldProps> = ({
 
   const renderDropdownSeparator = (): React.ReactNode => <View style={styles.dropdownSeparator} />;
 
-  const renderDropdownRow = (option: string, index: string, isSelected: boolean): React.ReactNode => {
-    let borderRadiusStyle = {};
-    const lastIndex = (options?.length ?? 0) - 1;
+  const renderDropdownRow = (option: string, index: any, isSelected: boolean): React.ReactNode => {
```
abubuwe, line 96: If Index is actually a number, why not specify the type?

selvach: Not 100% sure of myself - but because the type def of the source repo is wrong, if you put number, you signature of redderDropdownRow can't be passed in as a prop to the ModalDropdown from 'react-native-modal-dropdown';

PR #456, opened 2020-06-23
Create patient coordinator

the-kenneth:
# Description

Adds a patient coordinator to help manage the navigation and props in the patient flow. 

There is now a warning that the singletons have a require cycle, it works fine in this case. However this can be sorted by next adding in a parent coordinator and injecting dependencies. But this works for now.


src/CovidApp.tsx
```diff
@@ -38,7 +38,6 @@ import { ResetPasswordConfirmScreen } from '@covid/features/password-reset/Reset
 import { ResetPasswordScreen } from '@covid/features/password-reset/ResetPassword';
 import AboutYouScreen from '@covid/features/patient/AboutYouScreen';
 import PreviousExposureScreen from '@covid/features/patient/PreviousExposure';
-import StartPatientScreen from '@covid/features/patient/StartPatient';
```
selvach, line 41: NICE! Bye by StartPatientScreen!

src/core/patient/PatientCoordinator.ts
```diff
@@ -0,0 +1,84 @@
+import { StackNavigationProp } from '@react-navigation/stack';
+
+import { PatientStateType } from '@covid/core/patient/PatientState';
+import UserService from '@covid/core/user/UserService';
+import assessmentCoordinator from '@covid/core/assessment/AssessmentCoordinator';
+import { assessmentService } from '@covid/Services';
+
+import { ScreenParamList } from '../../features/ScreenParamList';
```
selvach, line 8: Use relative import here.

src/core/patient/PatientCoordinator.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.userService = userService;
+  };
+
+  startPatient = () => {
+    const { currentPatient } = this.patientData;
+    const config = this.userService.getConfig();
+    const patientId = currentPatient.patientId;
+
+    const startPage = 'WelcomeRepeat';
+    const shouldAskStudy = config.enableCohorts && currentPatient.shouldAskStudy;
```
selvach, line 56: Improved readability! +1

the-kenneth:
Had to make some changes in order to get the tests running. I am now passing the navigator (renamed to AppCoordinator) as a dependancy. This makes it so the child coordinators don't require each other. The AppCoordinator needs to be properly refactored though.

abubuwe:
Random comment - I don't think the Coordinators should live underneath the features folder. Arguably need their own space :)

PR #458, opened 2020-06-23
Add scroll view to us study invite

the-kenneth:
# Description

Quick fix for bug when using large accessibility font text.


the-kenneth:
I think modals are fine tbh. We should probably rework our own modal component though.

PR #479, opened 2020-06-29
Features/hide stats module in viral thankyou

snack-0verflow:
# Description

Remove stats module if country details is not valid.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 11 - 2020-06-29 at 15 39 40

## Checklist

- [x] I have run `npm test`


src/components/Cards/ShareCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { colors } from '@theme';
+
+import { RegularText } from '../Text';
+import { BrandedButton } from '../BrandedButton';
+
+interface Props {
+  area: AreaStatsResponse | null;
+}
+
+const shareUrl = i18n.t('share-this-app.url');
```
selvach, line 18: I'm suspicious this i18n string doesn't work when defined at the module level. It'll be evaluated before i18n.locale is set.

snack-0verflow: ahhhhh i see

PR #492, opened 2020-07-02
Fixes selection of mechanism other

selvach:
# Description

Where the Covid Test mechanism dropdown is set to "other" and then the user specifies free text, this free text is stored in the same field as the dropdown options (which should have been a typed enum). 

This then breaks, when you goto edit the Covid Test...

This PR fixes that by checking the mechanism value is one of the valid values for the dropdown. 

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Screenshot 2020-07-02 at 10 52 00

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/features/assessment/fields/CovidTestMechanismQuesion.tsx
+++ b/src/features/assessment/fields/CovidTestMechanismQuesion.tsx
@@ -1,15 +1,14 @@
 import { FormikProps } from 'formik';
 import React from 'react';
 import * as Yup from 'yup';
-import { ImageSourcePropType } from 'react-native';
 
 import i18n from '@covid/locale/i18n';
-import { FieldWrapper } from '@covid/components/Screen';
 import DropdownField from '@covid/components/DropdownField';
 import { LifestyleRequest } from '@covid/core/assessment/dto/LifestyleRequest';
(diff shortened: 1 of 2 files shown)


src/features/assessment/fields/CovidTestMechanismQuesion.tsx
```diff
@@ -91,9 +91,32 @@ export const CovidTestMechanismQuestion: CovidTestMechanismQuestion<Props, Covid
...
+  const valid_options = [
+    'nose_throat_swab',
+    'nose_swab',
+    'throat_swab',
+    'spit_tube',
+    'blood_sample',
+    'blood_sample_finger_prick',
+    'blood_sample_needle_draw',
+    'other',
+  ];
```
the-kenneth, line 103: Can this be turned into a enum and then also used in the `mechanismItems` above

selvach: Great suggestion! I also did the same for the TrainedWorker options. 

Not sure if the code for checking if a string is a valid member of a string enum is the best? @snack-0verflow @the-kenneth ?

the-kenneth:
Enum looks great!

PR #494, opened 2020-07-02
Add event when notification is pressed

the-kenneth:
# Description

This fires for any notifcation that is tapped on from out app. As we only have 1 for now this will work as expected. If we ever add more we will need to change this.


selvach:
Is the Description on the PR still valid?

the-kenneth:
> Is the Description on the PR still valid?

Description updated

PR #519, opened 2020-07-07
Fix splash screen issue.

selvach:
# Description

Fix splash screen issue. 
Call to `await appCoordinator.init(this.props.navigation as NavigationType)` can throw an error and so you can get stuck on the SplashScreen. 

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/features/SplashScreen.tsx
+++ b/src/features/SplashScreen.tsx
@@ -64,8 +64,12 @@ export class SplashScreen extends Component<Props, SplashState> {
   }
 
   async initAppState() {
-    await appCoordinator.init(this.props.navigation as NavigationType);
-    appCoordinator.gotoNextScreen(this.props.route.name);
+    try {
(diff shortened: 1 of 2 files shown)


src/features/SplashScreen.tsx
```diff
@@ -64,8 +64,12 @@ export class SplashScreen extends Component<Props, SplashState> {
   }
 
   async initAppState() {
-    await appCoordinator.init(this.props.navigation as NavigationType);
-    appCoordinator.gotoNextScreen(this.props.route.name);
+    try {
+      await appCoordinator.init(this.props.navigation as NavigationType);
```
abubuwe, line 68: how would the app coordinator be initialised after the error state?

PR #540, opened 2020-07-09
Disable Fastlane CI Step

abubuwe:
Disables the build + publish to App Center. Unnecessarily bloats the release process, given that no one is using App Center in the team.


abubuwe:
> Guess this is fine as we are mostly doing OTA releases

Yeah there is a larger task of reviewing our CI pipeline to ensure it still fully meets our needs

PR #548, opened 2020-07-13
bug: set default patient profile before starting diet study

snack-0verflow:
# Description

Make API to fetch and set first patient profile when user selected Diet Study from Drawer menu

TODO: Show user there is no profiles?

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I have run `npm test`

### DIFF
--- a/src/core/diet-study/DietStudyCoordinator.ts
+++ b/src/core/diet-study/DietStudyCoordinator.ts
@@ -47,8 +47,21 @@ export class DietStudyCoordinator {
     this.userService = userService;
   };
 
-  startDietStudy = () => {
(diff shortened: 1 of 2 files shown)


the-kenneth:
Isnt this already fetched in the init() call?

snack-0verflow:
> Isnt this already set in the init() call?

I believe the coordinator init call can be trigger with `currentPatient` as null. `currentPatient` aren't set until profile is selected. 

Before Diet study, all drawer menu item just opens a url so that was never an issue.

PR #560, opened 2020-07-15
Update end of study nav

the-kenneth:


src/core/diet-study/DietStudyCoordinator.ts
```diff
@@ -50,17 +50,16 @@ export class DietStudyCoordinator {
...
+        NavigatorService.navigate('DietStudyAboutYou', this.dietStudyParam); // Goes back to screen currently on stack
+      } else {
+        NavigatorService.navigate('DietStudyThankYou', this.dietStudyParam);
       }
-
-      NavigatorService.navigate('DietStudyThankYou', this.dietStudyParam);
     },
     DietStudyThankYou: () => {
-      NavigatorService.navigate('WelcomeRepeat');
+      NavigatorService.popTo('DietStudyIntro', true);
```
abubuwe, line 62: this looks like it goes from ThankYou to intro? Is that correct or have I misunderstood?

the-kenneth: The 2nd param is an inclusive flag, so it pops to 1 before the mentioned screen with set to true. Did this because the study can either be started from the menu or the assessment flow. This pops back to where it was started from

PR #586, opened 2020-07-21
Fix select profile screen

the-kenneth:
# Description

Fix to make profiles load correctly. Removed some unnecessary code in profile hook.

### DIFF
--- a/src/features/multi-profile/ProfileList.hooks.ts
+++ b/src/features/multi-profile/ProfileList.hooks.ts
@@ -15,7 +15,6 @@ import appCoordinator from '../AppCoordinator';
 export const useProfileList = (navigation: DrawerNavigationProp<ScreenParamList, 'SelectProfile'>) => {
   const userService = useInjection<ICoreService>(Services.User);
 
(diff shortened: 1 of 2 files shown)


src/features/multi-profile/ProfileList.hooks.ts
```diff
@@ -15,7 +15,6 @@ import appCoordinator from '../AppCoordinator';
 export const useProfileList = (navigation: DrawerNavigationProp<ScreenParamList, 'SelectProfile'>) => {
   const userService = useInjection<ICoreService>(Services.User);
 
-  const [shouldRefresh, setShouldRefresh] = useState<boolean>(false);
```
snack-0verflow, line 18: The bug can be fix if we default `shouldRefesh` to true.

snack-0verflow: I was thinking about add `setShouldRefresh(false);` in `listProfiles` and `final { setShouldRefresh(true); }` to the try / catch block

the-kenneth: I was wondering what is shouldRefresh used for, I couldnt see anything tbh?

PR #600, opened 2020-07-24
Add profile postcode editing

the-kenneth:
# Description

Made changes around the select profile screen, profile list hook and how the edit profile is handled.

Now the patient info is loaded when edit profile is pressed and this can then be edited. I also simplified the profile list hook and moved the navigation out of it. 

Additionally I deleted the PatientProfile type which was a subset of Profile, and I just pass that around now. It was a confusing type.

There is 1 todo in the EditLocation screen, i couldnt figure out how to _easily_ push the content to the bottom of the screen. Any ideas welcome


src/features/multi-profile/edit-profile/EditLocationScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  formikProps={props}
+                  label={i18n.t('edit-profile.location.label')}
+                  placeholder={i18n.t('placeholder-postcode')}
+                  name="postcode"
+                  inputProps={{ autoCompleteType: 'postal-code' }}
+                  showError
+                />
+                {
+                  // TODO : How to push below to the bottom simply?
+                }
```
the-kenneth, line 77: This is the todo mentioned above

src/core/user/UserService.ts
```diff
@@ -333,10 +333,11 @@ export default class UserService extends ApiClientBase implements ICoreService {
       patientName = i18n.t('default-profile-name');
     }
 
-    const profile: PatientProfile = {
+    const profile: Profile = {
+      id: patientState.patientId,
       name: patientName,
-      avatarName: (patient.avatar_name || 'profile1') as AvatarName,
-      isPrimaryPatient: !patient.reported_by_another,
+      avatar_name: patient.avatar_name ? patient.avatar_name : DEFAULT_PROFILE,
```
snack-0verflow, line 339: `avatar_name: patient.avatar_name ?? DEFAULT_PROFILE` A different syntax, not suggesting changes. just sharing

src/features/multi-profile/edit-profile/EditLocationScreen.tsx
snack-0verflow: @the-kenneth Maybe flex box's attributes, `justifycontent` with `space-bewteen`? (Maybe)

the-kenneth: Played around with that, couldn't get it working. Not sue if its to do with <form> being embedded

PR #613, opened 2020-07-28
Fix handling of bad tokens and refactor CountrySelectScreen

selvach:
# Description

Commit: 0191416
- Removes the patientID as an optional param on the "CountrySelect". This was never used. 
- Forces a navigate to CountrySelect when userservice.logout() is called
- Adds missing awaits. 

Commit 8959d2b
- Removed duplicate useEffect on the DrawerMenu

Commit 0191416
 - This simplifies the userService.getProfile() method which is called in the constructor. 
- . It would always call the sever. If AsyncStorage was empty, it would return the server response. If AsyncStorage had a previous response stored it would return the cached version and then replace the AsyncStorage with response from the server. Which means you might get a stale response.... I think this is why people are getting stuck on Splash screen!
 - Now, if you get a 404 or 401 on getProfile to the server this function throws a retryable exception which displays a retry button on the splash screen

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


src/core/user/UserService.ts
```diff
@@ -411,27 +413,17 @@ export default class UserService extends ApiClientBase implements ICoreService {
...
+      .then(async (profileResponse) => {
+        await AsyncStorageService.saveProfile(profileResponse.data);
+        return profileResponse.data;
+      })
+      .catch(async (error) => {
+        await this.logout();
+      });
 
-    return localProfile;
+    return {} as UserResponse;
```
snack-0verflow, line 426: @selvach is this intentional?

selvach:
@snack-0verflow Made changes to the logic - I realised that forcing a logout if the getProfile() failed would log out anyone who opened the app without an internet connection. 

Instead I'm throwing a "RetryableException" that enables this on the splash screen:

Screenshot 2020-07-28 at 19 29 25

selvach:
Fixed merge conflicts @the-kenneth @snack-0verflow

PR #626, opened 2020-07-31
Edit about you

the-kenneth:
# Description

Add edit 'About You' page. Introduces a coordinator interface which allows the 'About You' page to be reused, currently switching coordinator by passing an 'editing' boolean to the screen. Im not 100% happy with this because I would much rather pass the the whole coordinator but this works.

Ive also updated the 'PatientData' object to contain the 3 pieces of patient data we use, 'Profile', 'State' and 'Info'. Ideally in the future we refactor all pages to just pass this one encompassing object but for now I started using it on the editing pages.

Ive currently left the edit location still there so users can find it easily, but postcode can be modified from either.


src/features/multi-profile/edit-profile/EditProfileScreen.tsx
```diff
@@ -19,29 +19,31 @@ type RenderProps = {
...
 
-        <TouchableOpacity
-          style={styles.profileLabel}
-          onPress={() => appCoordinator.goToEditLocation(profile, patientInfo)}>
+        <TouchableOpacity style={styles.profileLabel} onPress={() => editProfileCoordinator.goToEditLocation()}>
           <HeaderText>Your location</HeaderText>
           <Image style={styles.chevron} source={chevronRight} />
         </TouchableOpacity>
+
+        <TouchableOpacity style={styles.profileLabel} onPress={() => editProfileCoordinator.goToEditAboutYou()}>
```
selvach, line 37: Could you wrap this in a feature flag via the appConfig or countryConfig please? Looks like legal might have an issue with this feature. (Specifically showing and editing PII / Health data, without having some sort of passcode lock. Basically anyone with the phone could view health data). 

Then happy to merge!

PR #660, opened 2020-08-07
increase machine size

abubuwe:
increase size of CI machine to quash memory errors


abubuwe:
just holding back on merging this in case this is more a symptom than the actual cause of the problem

abubuwe:
Closing for now as it doesn't seem needed

PR #663, opened 2020-08-10
Features/personalised map card

snack-0verflow:
# Description

- Map card (Uses active cases from `startup_info.local_data`)
- Update center & zoom in using `startup_info.local_data.map_config` (if avaliable then center will be defaulted to UK and zoomed out)
- Full-screen map view
- Include carto map in empty state

CSS styling for embedded html:
- The legend (list of cases pm) hides for smaller height, this is for the embedded map in a card. 

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I've added analytics as relevent
- [x] I have run `npm test`

BE needs to return `lat` and `lng` for `startup_info.local_data.map_config`


snack-0verflow:
@abubuwe @the-kenneth @selvach

PR #665, opened 2020-08-11
Disable bottom tab nav

the-kenneth:
Replaces the component directly with the dashboard screen.


snack-0verflow:
@the-kenneth Looks good

PR #692, opened 2020-08-17
fix: add 'Dashboard' to nav stack when resetting

snack-0verflow:
# Description

Added `Dashboard` to nav stack while resetting, also set current nav index to `SelectProfile`

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [x] I have run `npm test`

### DIFF
--- a/src/features/thank-you/ThankYouUKScreen.tsx
+++ b/src/features/thank-you/ThankYouUKScreen.tsx
@@ -133,7 +133,7 @@ export default class ThankYouUKScreen extends Component<RenderProps, State> {
                   onPress={() =>
                     this.props.navigation.reset({
                       index: 1,
-                      routes: [{ name: 'Dashboard' }, { name: 'SelectProfile' }],
+                      routes: [{ name: appCoordinator.homeScreenName }, { name: 'SelectProfile' }],
                     })
                   }
                   style={styles.ctaMultipleProfileText}>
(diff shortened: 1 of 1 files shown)


src/features/thank-you/ThankYouUKScreen.tsx
```diff
@@ -132,8 +132,8 @@ export default class ThankYouUKScreen extends Component<RenderProps, State> {
                 <ClickableText
                   onPress={() =>
                     this.props.navigation.reset({
-                      index: 0,
-                      routes: [{ name: 'SelectProfile' }],
+                      index: 1,
+                      routes: [{ name: 'Dashboard' }, { name: 'SelectProfile' }],
```
abubuwe, line 136: Not 100% on how the screen naming works, but does this cover new + old homescreens (is the old one still called WelcomeRepeat)

snack-0verflow: True. Will update this

snack-0verflow: @abubuwe Used `appCoordinator.homeScreenName` to keep it dynamic

abubuwe: Can you confirm if this is needed to be done anywhere else with a quick text search?

snack-0verflow: Did a global search of `reset(` I believe this is the only place that requires a default home screen when resetting the stack. 

Other reset stack method does include a base route & or user is logged out.

PR #698, opened 2020-08-18
fix: track insights event if country code is different

snack-0verflow:
# Description

- Show country picker on app start if `country_code` from API is different to one saved locally.
- Update `country_code` using Covid app BE API
- Call API to update user's country code

UPDATE:
- Disable UI popup
- Added analytics event to track coutnry code mismatch

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I've added analytics as relevent
- [x] I have run `npm test`

Require BE work on updating user's country code.


snack-0verflow:
@abubuwe

src/core/user/UserService.ts
```diff
@@ -166,6 +167,17 @@ export default class UserService extends ApiClientBase implements IUserService {
     return null;
   }
 
+  public async updateCountryCode(body: UpdateCountryCodeRequest): Promise<UserResponse | null> {
+    try {
+      const { data } = await this.client.patch<UserResponse>(`/users/country_code/`, body);
+      await AsyncStorageService.saveProfile(data);
```
selvach, line 173: saveProfile() has been deleted from AsyncStorageService in this PR. 
I didn't see any point in storing the response. 
https://github.com/zoe/covid-tracker-react-native/pull/708

src/core/user/dto/UserAPIContracts.ts
```diff
@@ -65,6 +65,11 @@ export type UserResponse = {
   patients: string[];
   ask_for_rating: boolean;
   is_tester: boolean;
+  country_code: string;
+};
+
+export type UpdateCountryCodeRequest = {
+  country_code: string;
```
selvach, line 72: You could make this into an Enum (GB, US, SE)

snack-0verflow:
@selvach Made some changes. Added enums & removed redundant code

src/features/AppCoordinator.ts
```diff
@@ -117,14 +130,30 @@ export class AppCoordinator {
   };
 
   async init() {
+    let shouldShowCountryPicker = false;
+    const profile: UserResponse | null = null;
```
selvach, line 134: Change line 134: from const -> let
Remove the const in line 140. Otherwise you are hiding the scope and line 144 will never evaluate as true.

selvach (changes requested):
.

src/features/AppCoordinator.ts
snack-0verflow: @selvach my bad. just pushed a commit up

selvach (approved):
Thanks for the changes! Tested locally and it looks good.

PR #705, opened 2020-08-19
Fix US Study Modal

selvach:
# Description

Fixes a problem when dismissing a the US Study Invite Modal. (Only shown for new US users). 

This was a caused by a circular reference between between ConsentService and PatientService. 

The USStudyInvite "consent" is an anomoly - as the flag is on the patient model. So `setUSStudyInviteResponse` really does belong in the PatientService. 

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device


abubuwe (approved):

abubuwe:
Thanks for jumping on this so quickly 👌

PR #729, opened 2020-08-24
feat: enable postcode edit from map

snack-0verflow:
# Description

- Enable edit of property `postcode` for the main Patient

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Checklist

- [x] I have run `npm test`


abubuwe:
This GIF highlights the fact that even when a postcode is updated, we don't try to refetch the map at present, and we may want to, otherwise users will think their email still requires updating?

abubuwe (approved):

PR #733, opened 2020-08-26
Patientdata refactor

the-kenneth:
# Description

Refactor all flows to use and pass patientData object instead of a different combination of patientState, patientInfo and profile for each one. This then in turn tidies up the patientService as you now just request for the patientData object rather than the individuals. Ive also renamed some functions to properly match their returned object name.


the-kenneth:
@selvach @snack-0verflow 

Dont forget this :)

snack-0verflow (approved):

PR #735, opened 2020-08-27
Fix/Refresh homescreen data on Login

abubuwe:
# Description

Triggers a refetch of startupInfo on Login to ensure any returning users, or users who've logged out for any reason still see the dashboard.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

### DIFF
--- a/src/features/AppCoordinator.ts
+++ b/src/features/AppCoordinator.ts
@@ -136,7 +136,6 @@ export class AppCoordinator {
     let profile: UserResponse | null = null;
 
     await this.userService.loadUser();
-    const info = await this.contentService.getStartupInfo();
 
     if (this.userService.hasUser) {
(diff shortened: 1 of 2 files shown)


selvach (approved):

selvach:
I've just added `show_edit_location` on the startup_info API response on the backend. Could you add to conditionally show Edit Postcode on the homescreen?  I think we can leave it on the Profile Info screen to 100% if people stumble across it.

abubuwe:
> I've just added `show_edit_location` on the startup_info API response on the backend. Could you add to conditionally show Edit Postcode on the homescreen? I think we can leave it on the Profile Info screen to 100% if people stumble across it.

Done in a separate PR.

PR #755, opened 2020-09-08
Feature/time test taken

petermason:
# Description

Add time test take to NHS study form
select-time-component
select-time-component-error
select-time-component-modal

Has custom button to replicate design (See Figma) (ActionButton)


abubuwe:
Basing purely off of screenshot but it appears to say "Pick a date" instead of "Pick a time"

src/features/covid-tests/fields/CovidTestDateQuestion.tsx
```diff
@@ -38,6 +38,7 @@ export const CovidTestDateQuestion: CovidTestDateQuestion<Props, CovidTestDateDa
   const [state, setState] = useState({
     showDatePicker: false,
     showRangePicker: false,
+    showTimePicker: false,
```
selvach, line 41: I think this line is not required?

petermason:
> Basing purely off of screenshot but it appears to say "Pick a date" instead of "Pick a time"

fixed and translations updated

src/features/covid-tests/fields/CovidTestDateQuestion.tsx
petermason: fixed, removed redundant code.

selvach:
Testing this locally the UI looks great! But I'm having problems submitting:

1) I get the following problem on the validation, even after selecting the date:

Screenshot 2020-09-08 at 14 28 04

2) These lines indicate we're probably not sending the value to the backend? 

```
CovidTestTimeQuestion.createDTO = (formData: ICovidTestTimeData): Partial<CovidTest> =>
  ({
    time_taken: '',
  } as Partial<CovidTest>);
```

You'll need to take `formData.dateTestTime` (date) and cast to the string format required by the backend.

selvach:
Tested locally and LGTM! I can see the data flowing to the backend and persisted.

selvach:
Before we roll this to a different country (which we currently don't plan to), we'll have to start capturing user timezone information. The backend is converting to UTC!

selvach (approved):

PR #762, opened 2020-09-10
Update privacy and consent copy

the-kenneth:


abubuwe:
can you update the privacy and consent version numbers please

PR #804, opened 2020-09-28
Remove OptionalInfo and dont show YourStudy in registration flow

the-kenneth:


src/features/register/gb/PrivacyPolicyUKScreen.tsx
```diff
@@ -83,8 +83,7 @@ export class PrivacyPolicyUKScreen extends Component<PropsType, object> {
 
           <RegularText>
             We also collect contact information and other information from your device including:
-            {'\n\n'}- Your name (optional){'\n'}- Email address (optional){'\n'}- Phone number (optional){'\n'}- A user
-            name and password{'\n'}- IP address{'\n'}- Device ID{'\n'}
+            {'\n\n'}- Email address (optional){'\n'}- A user name and password{'\n'}- IP address{'\n'}- Device ID{'\n'}
```
selvach, line 86: Ah... let's revert this change in Privacy Policy. Defo, work doing, now we're not collecting this - but it requires generating a new PDF and hosting on the website. I'll check with Christina and Francis (legal).

selvach (approved):

src/features/register/gb/PrivacyPolicyUKScreen.tsx
the-kenneth: I just checked and she confirmed you are right!

PR #805, opened 2020-09-29
Fix bug that can return user to incorrect screen after ThankYou screens

the-kenneth:
ThankYou screens were still using old nav code which could sometimes take the user to the wrong screens. e.g. after first assessment


the-kenneth:
Test gives an unhelpful error and fails when I extended the AssessmentCoordinator from Coordinator. Not sure what is wrong tbh.

the-kenneth:
Any ideas? @selvach @snack-0verflow

selvach (approved):

PR #807, opened 2020-09-29
Add OptionalInfoScreen back

selvach:
This PR partially reverts PR #804 after Richard pointed out that this info was used by NHS Wales and some MGH Cohorts. StudyScreen can still be removed however.


src/core/patient/PatientCoordinator.ts
```diff
@@ -22,12 +22,6 @@ export class PatientCoordinator extends Coordinator implements UpdatePatient {
   private readonly localisationService: ILocalisationService;
 
   screenFlow: Partial<ScreenFlow> = {
-    NHSIntro: () => {
```
the-kenneth, line 25: Why have NHS screens been removed?

selvach: So the NHS screens are only shown if `this.patientData.patientState.isNHSStudy` is True... but this can only be set from the Studies screen, which is now removed from the flow. So no-one will see this option when signing up / creating a profile. The only way to access this is via Edit Profile -> My Studies -> tick NHS Study -> get shown the NHS Intro / NHSDetails screens...

Screenshot 2020-09-30 at 11 43 40

the-kenneth (approved):

PR #819, opened 2020-10-08
We need to be able to receive notifications after a reboot!

andyguest:
# Description

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


andyguest:
Inviting some opinions from the team: will this change fix our problems without introducing other problems for people on upgrade, e.g. if we end up requesting additional permissions and this causes friction / customer support impact?

It is my reading of the docs that "android.permission.RECEIVE_BOOT_COMPLETED" is considered a normal-level permission which no Android system would bother to prompt the user for (because the user cannot say no!).

See: https://developer.android.com/guide/topics/permissions/overview#normal-dangerous
And the reference list here: https://developer.android.com/reference/android/Manifest.permission

the-kenneth:
Put in review im good to try it

andyguest:
@the-kenneth I've done no build / no testing, except to verify that the existing app exhibits a failure to receive notifications after a restart (using the expo tester here https://expo.io/notifications against my personal expo token).  Is this something you would take over on my behalf?

the-kenneth:
I tested with this on and off and in both situations I was able to receive notifications after restarting even when not reopening the app. It did take a few mins sometimes to get them after a reset but they did all come through.

PR #833, opened 2020-10-15
Fix incorrect margins on dashboard and icon resolution

the-kenneth:
Removes grey block at bottom of dashboard screens and increases resolution of the logo as it was blury.


selvach:
Thank you for new binary file ... that was really bothering me. 

Unfortunately the margins look a bit squiff on my device... see attached:

Screenshot_20201015-210812
Screenshot_20201015-210816

the-kenneth:
Looks like the startup info didn't load for you and the contributor numbers are missing. However I have tightened the top section anyway to match what it looked like before. Give it a go again.

selvach (approved):

PR #834, opened 2020-10-15
Fix disabling scroll on screen for Android devices

the-kenneth:


selvach:
Looks great - tested the COVID Test List Screen.

selvach (approved):

PR #836, opened 2020-10-15
Update blog assets

selvach:
Blog Post:
- Add Blog post link to "Can You Catch COVID Twice"
- Add optional imageStyles so we can pass in a border radius as we don't have Nils around to create a proper asset. 

Video Link
- Add Tim's latest youtube video link
- I'm using an older thumbnail with the date cropped from the image. I've added the text underneath. 

Screenshot 2020-10-15 at 22 34 23


snack-0verflow (approved):
.

PR #838, opened 2020-10-16
Fix blood group order and minimum dropdown height

the-kenneth:
Puts 'I dont know' to the top and then adds 20px to the minimum drop down height in order to partially expose any additional items.


andyguest (commented):
Sounds like a great fix with the height change.  My only concern with the reordering would be if it encourages laziness and risks reducing the quality of data for what many people believe to be potentially a blood disease with other impacts.

the-kenneth:
@andyguest reverted the order

andyguest (approved):
LGTM.  Not sure if we have captured the pixel maths you did in any comments - might be good to consider adding in the future.

PR #858, opened 2020-10-28
Only allow secondary profiles under 20 to join school network

the-kenneth:
Roughly estimate age to limit which accounts can join a school network


src/features/multi-profile/edit-profile/EditProfileScreen.tsx
```diff
@@ -53,10 +53,11 @@ export const EditProfileScreen: React.FC<RenderProps> = (props) => {
         )}
 
         {editProfileCoordinator.shouldShowSchoolNetwork() && (
-          <>
```
andyguest, line 56: what was this line?

src/features/multi-profile/edit-profile/EditProfileCoordinator.ts
```diff
@@ -98,6 +98,19 @@ export class EditProfileCoordinator extends Coordinator implements UpdatePatient
...
   shouldShowSchoolNetwork() {
+    const currentPatient = this.patientData.patientState;
+    const birthYear = this.patientData.patientInfo?.year_of_birth;
+
+    // If undefined - the patient hasn't completed the register flow
+    if (!birthYear) {
+      return false;
+    }
+    const age = new Date().getFullYear() - birthYear;
+    const schoolAged = age > 1 && age < 20;
```
andyguest, line 109: Harsh on the kids of today! :-)

andyguest (approved):
This looks highly promising. Might be best to get a second opinion though :-)

src/features/multi-profile/edit-profile/EditProfileScreen.tsx
selvach: This is an empty View element - which is required to contain multiple `LinkItem` children.

PR #870, opened 2020-11-05
Fix margins for height and weight questions

the-kenneth:


selvach:
Could you include a screen shot next time, please! 
Tested and it looks good!

Screenshot 2020-11-05 at 13 14 05

selvach (approved):

PR #874, opened 2020-11-06
Disable features for minors

the-kenneth:
The YourWork screen has been disabled for minors and the last 4 questions of AboutYou also get hidden if the entered age is a minor.

### DIFF
--- a/src/core/patient/PatientService.ts
+++ b/src/core/patient/PatientService.ts
@@ -165,6 +165,9 @@ export class PatientService extends ApiClientBase implements IPatientService {
     const isNHSStudy = patient.is_in_uk_nhs_asymptomatic_study;
     const hasSchoolGroup = patient.has_school_group;
 
+    const age = new Date().getFullYear() - patient.year_of_birth;
+    const isMinor = age > 0 && age < 20;
+
     return {
       ...patientState,
       profile,
@@ -186,6 +189,7 @@ export class PatientService extends ApiClientBase implements IPatientService {
       hasBloodGroupAnswer,
       isNHSStudy,
       hasSchoolGroup,
+      isMinor,
     };
   }
 
--- a/src/core/patient/PatientService.ts
+++ b/src/core/patient/PatientService.ts
@@ -4,14 +4,13 @@ import i18n from '@covid/locale/i18n';
 import { Services } from '@covid/provider/services.types';
 import { DEFAULT_PROFILE } from '@covid/utils/avatar';
 import { isUSCountry, isGBCountry } from '@covid/core/localisation/LocalisationService';
-import { getDaysAgo } from '@covid/utils/datetime';
 import { PatientInfosRequest } from '@covid/core/user/dto/UserAPIContracts';
 import { IConsentService } from '@covid/core/consent/ConsentService';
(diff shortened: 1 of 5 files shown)


src/core/patient/PatientService.ts
```diff
@@ -165,6 +164,8 @@ export class PatientService extends ApiClientBase implements IPatientService {
     const isNHSStudy = patient.is_in_uk_nhs_asymptomatic_study;
     const hasSchoolGroup = patient.has_school_group;
 
+    const isMinor = isMinorAge(patient.year_of_birth);
```
selvach, line 167: Could this be null / non-numeric?

src/features/patient/AboutYouScreen.tsx
```diff
@@ -127,6 +128,7 @@ export default class AboutYouScreen extends Component<AboutYouProps, State> {
           currentPatient.isFemale = formData.sex !== 'male';
           currentPatient.isPeriodCapable =
             !['', 'male', 'pfnts'].includes(formData.sex) || !['', 'male', 'pfnts'].includes(formData.genderIdentity);
+          currentPatient.isMinor = isMinorAge(cleanIntegerVal(formData.yearOfBirth));
```
selvach, line 131: Just checking if cleanIntegerVal needs a null check?

src/features/patient/AboutYouScreen.tsx
```diff
@@ -428,29 +432,33 @@ export default class AboutYouScreen extends Component<AboutYouProps, State> {
...
-                    selectedValue={props.values.helpAvailable}
-                    onValueChange={props.handleChange('helpAvailable')}
-                  />
-
-                  <YesNoField
-                    label={i18n.t('mobility-aid')}
-                    selectedValue={props.values.mobilityAid}
-                    onValueChange={props.handleChange('mobilityAid')}
-                  />
+                  {!isMinor && (
```
selvach, line 435: Payload submitted to the backend contains `false` for these values:

```
    "housebound_problems": false,
    "needs_help": false,
    "help_available": false,
    "mobility_aid": false,
```

Which is better: `null` or assume `false`?

src/features/patient/AboutYouScreen.tsx
the-kenneth: The form validation would not allow a submit if this is not correctly filled, thus you couldn't hit this code path.

src/features/patient/AboutYouScreen.tsx
the-kenneth: My assumption was set as false, this means the current data set wouldnt have a new value introduced. I can check double check this.

src/core/patient/PatientService.ts
the-kenneth: This should never be null or non-numeric as a submit to the backend is validated previously. Am I correct in thinking that?

src/features/patient/AboutYouScreen.tsx
the-kenneth: Ive made changes to not include those questions into the payload if not answered.

selvach (approved):

PR #878, opened 2020-11-09
Dont use assessment screen when selecting edit profile

the-kenneth:
This is not an ideal solution but it works. I spent quite a while looking to make it better but decided the refactoring would be too much.

The coordinator pattern is something that is very helpful for us IMO but it may need slight tweaking in future projects to get this use case to work smoothly. (Using a screen independently in 2 different flows and having the buttons call backs handled by the coordinator).

Ideally I would want to pass coordinators to a screen so it is totally decoupled from logic, however, even though this can be done, by passing a non-serializable object it has potential side effects with rebuilding state (if we were to ever do that) and also deep linking. The code in its current state might not need too much refactoring to get something like this working but I would want to discuss potential options if we do want to explore this for future learnings.

### DIFF
--- a/src/features/AppCoordinator.ts
+++ b/src/features/AppCoordinator.ts
@@ -118,7 +118,7 @@ export class AppCoordinator extends Coordinator implements SelectProfile, Editab
       NavigatorService.navigate('SelectProfile', { assessmentFlow: true });
     },
     ArchiveReason: () => {
-      NavigatorService.navigate('SelectProfile', { assessmentFlow: true });
+      NavigatorService.navigate('SelectProfile'); // Go back to SelectProfile with last used params
     },
     ValidationStudyIntro: () => {
       NavigatorService.navigate('ValidationStudyInfo');
--- a/src/features/AppCoordinator.ts
+++ b/src/features/AppCoordinator.ts
@@ -34,7 +34,7 @@ import {
 } from '@covid/core/content/state/contentSlice';
 import { ScreenParamList } from '@covid/features/ScreenParamList';
 import { UserResponse } from '@covid/core/user/dto/UserAPIContracts';
-import { Coordinator, SelectProfile } from '@covid/core/Coordinator';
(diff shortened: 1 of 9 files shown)


src/features/AppCoordinator.ts
```diff
@@ -108,17 +108,17 @@ export class AppCoordinator extends Coordinator implements SelectProfile {
...
       }
     },
     Dashboard: () => {
       // UK only so currently no need to check config.enableMultiplePatients
-      NavigatorService.navigate('SelectProfile', { editing: true });
+      NavigatorService.navigate('SelectProfile', { assessmentFlow: true });
     },
     ArchiveReason: () => {
-      NavigatorService.navigate('SelectProfile', { editing: true });
+      NavigatorService.navigate('SelectProfile', { assessmentFlow: true });
```
selvach, line 121: This is minor.... But:

- Click DrawerMenu -> Edit Profile (asssesmentFlow = False) 
- Choose a Profile
- Archive that profile
- Now you return to the SelectProfile Screen by assessmentFlow = False

selvach (approved):
I like the approach. Keep it simple - this PR is also nice an easy to review. 

Testing looks good - there's a minor bug. Do you want to fix?

selvach:
Screenshot 2020-11-10 at 09 31 33

I wonder if this screen needs a "Back to Homescreen" link? 
After you finish editing a profile, you have to hit the back button multiple times, which feels a bit clunky. Could you check with Mathias?

the-kenneth:
> Screenshot 2020-11-10 at 09 31 33
> 
> I wonder if this screen needs a "Back to Homescreen" link?
> After you finish editing a profile, you have to hit the back button multiple times, which feels a bit clunky. Could you check with Mathias?

Maybe this can be a separate issue? Ticket scope doesn't cover it :)

selvach (approved):

PR #916, opened 2020-11-18
Stage

snack-0verflow:
# Description

- install `expo-notifications`
- updated code in `App.tsx` to handle on notification interact
- @selvach updated code in `expo.ts` to specify an experenice id for expo push notification
- copy `google-services.json` to `/android/app`
- turned on push notification capability on iOS

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "expo-font": "^8.3.0",
     "expo-intent-launcher": "~8.3.0",
     "expo-localization": "~9.0.0",
+    "expo-notifications": "~0.7.2",
     "expo-sharing": "~8.4.1",
     "expo-splash-screen": "~0.6.1",
     "expo-status-bar": "~1.0.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "Covid",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "scripts": {
     "start": "react-native start",
     "start:reset": "react-native start --reset-cache",
--- a/src/core/push-notifications/expo.ts
+++ b/src/core/push-notifications/expo.ts
@@ -1,5 +1,5 @@
 import Constants from 'expo-constants';
-import { Notifications } from 'expo';
+import * as Notifications from 'expo-notifications';
 import * as Permissions from 'expo-permissions';
 
 import { IPushTokenEnvironment } from './PushNotificationService';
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "expo-font": "^8.3.0",
     "expo-intent-launcher": "~8.3.0",
     "expo-localization": "~9.0.0",
+    "expo-notifications": "~0.7.2",
```
andyguest, line 66: Just asking the dumb question: why don't you also need an expo-application package here?

src/core/push-notifications/expo.ts
```diff
@@ -13,7 +14,9 @@ export default class ExpoPushTokenEnvironment implements IPushTokenEnvironment {
     let token = null;
     try {
       if (await this.isGranted()) {
-        token = await Notifications.getExpoPushTokenAsync();
+        const { data } = await Notifications.getExpoPushTokenAsync({ experienceId: '@julien.lavigne/covid-zoe' });
```
andyguest, line 17: Where should this internal knowledge of "experienceId" come from?  It shouldn't be hardwired in this method call, although it could reasonably (for now) be a member on the ExpoPushTokenEnvironment object.  This is how we inadvertently create untestable code.

yarn.lock
```diff
@@ -6589,7 +6594,7 @@ expo-asset@~8.2.0:
     path-browserify "^1.0.0"
     url-parse "^1.4.4"
 
-expo-constants@~9.2.0:
+expo-constants@9.2.0, expo-constants@~9.2.0:
```
andyguest, line 6597: Why did this change happen?  Looks wrong to me.

andyguest (commented):
A few questions.

package.json
snack-0verflow: We can add it if we want to get app info such ID, app name, and build. `expo-application` would be one of the available packages to use to fetch that info

src/core/push-notifications/expo.ts
snack-0verflow: `app.json` this has expo configs

package.json
snack-0verflow: I see what you mean. is in the `yarn.lock`

snack-0verflow: I believe is because `expo-application` is a peer dependency for something else. Hence why we don't have to include it ourselves

yarn.lock
snack-0verflow: This is generated by the package manager

package.json
andyguest: it's not because we haven't done the iOS testing yet then? :-)

andyguest (dismissed):
I think Chuck has survived my grilling comfortably.

selvach (approved):

selvach:
FYI - We created a new Firebase project for stage and add the google-services.json for that new stage project into CircleCI

PR #939, opened 2020-11-24
Force new push notifications

selvach:
A users PushToken will change when:
 - Uninstalling and installing the app
 - Clearing AppData

https://forums.expo.io/t/push-token-changed-after-which-event/2209/3

On opening the Dashboard, `refreshPushNotifcaion()` is called. 

- If a token is not found in AsyncStorage - generate a new ExpoToken using `Notifications.getExpoPushTokenAsync()` and send it to the backend. 
- If a token is found in AsyncStorage and it's older than a week, then send it to the backend - but it uses the value found in AsyncStorage. And this will be wrong for the 2 of the cases above. 
- Otherwise, do nothing. 

This PR changes two things:
 - Always call Notifications.getExpoPushTokenAsync() rather than use the one on AsyncStorage, which maybe out of invalid.


snack-0verflow (approved):

snack-0verflow (dismissed):

andyguest (commented):
It looks like the combination of this fix and a previous change to remove old push tokens from async storage will mean that we can resolve the problems some of our users have had immediately by asking them to uninstall and reinstall (which is the approach that the Support team take).

For people who don't get in touch with Support, I presume that we will actually take up to one week to determine that we need to refresh their stale tokens, but everything should be ok for them after that has happened.

andyguest:
My remaining concern is whether this will result in an explosion of pushTokens in the database because we don't remove the old unusable ones at any point.  Are we ever going to have to address this?  I guess we've got plenty of time ahead to do so...

bionara (approved):
Code looks good

PR #957, opened 2020-12-01
Switch to dashboard for US users

the-kenneth:
Simulator Screen Shot - iPhone 11 - 2020-12-01 at 10 03 00


selvach:
Looks fab - I've tested on iOS Simulator!

selvach (approved):

PR #964, opened 2020-12-01
Add a minimum threshold of 3

selvach:
Add a minimum threshold of 3 on the front end display when computing the number of parents to be returned.


src/components/Cards/school-networks/SchoolStats.tsx
```diff
@@ -18,14 +18,16 @@ interface IProps {
 }
 
 const GROUP_ACTIVATION_THRESHOLD = 0.1;
+const GROUP_ACTIVATION_MINIMUM = 3;
 
 function SchoolStats({ active, bubbleName = '', daily = undefined, isLast = false, reported, size, total }: IProps) {
   const getToSignUpMessage = () => {
     const count = Math.round(total * GROUP_ACTIVATION_THRESHOLD) - size;
```
andyguest, line 25: I'm confused again.  Why are we repeating backend logic here?  Shouldn't we be able to rely on the active flag from the bubble?

andyguest: oh, I see.  we should return this "number needed" from the endpoint really.  never mind.

andyguest (approved):

src/components/Cards/school-networks/SchoolStats.tsx
selvach: Yah - exactly! 

I thought about returning the "number needed" from the backend, so we didn't need to spread this logic around. But made a trade off. This was added into the design a little late...

PR #974, opened 2020-12-03
test for user

petermason:
# Description

FIXES:

from homescreen, log out
land on the 3 flags page
PRESS THE BACK BUTTON
Relaunch the app, either directly to the homescreen (on a Samsung A50) or back to 3 flags on my OnePlus 6T.
if relaunched to 3 flags, press the back button.
if app exits, relaunch it
=> HOME SCREEN!

- [X] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/core/user/UserService.ts
+++ b/src/core/user/UserService.ts
@@ -75,7 +75,6 @@ export default class UserService extends ApiClientBase implements IUserService {
   public async logout() {
     this.hasUser = false;
     await this.deleteLocalUserData();
-    // NavigatorService.navigate('CountrySelect');
   }
 
   private async deleteLocalUserData() {
--- a/src/core/user/UserService.ts
+++ b/src/core/user/UserService.ts
@@ -75,7 +75,7 @@ export default class UserService extends ApiClientBase implements IUserService {
   public async logout() {
     this.hasUser = false;
     await this.deleteLocalUserData();
-    NavigatorService.navigate('CountrySelect');
+    // NavigatorService.navigate('CountrySelect');
   }
 
   private async deleteLocalUserData() {
--- a/src/features/AppCoordinator.ts
+++ b/src/features/AppCoordinator.ts
@@ -68,6 +68,8 @@ export class AppCoordinator extends Coordinator implements SelectProfile, Editab
 
   shouldShowCountryPicker: boolean = false;
 
+  thisUser: UserResponse | null;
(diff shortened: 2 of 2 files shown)


src/core/user/UserService.ts
```diff
@@ -75,7 +75,7 @@ export default class UserService extends ApiClientBase implements IUserService {
   public async logout() {
     this.hasUser = false;
     await this.deleteLocalUserData();
-    NavigatorService.navigate('CountrySelect');
```
selvach, line 78: Did you mean to leave this commented out?

petermason: i actually meant to remove it! It gets called in the Logout.hooks.ts file

selvach (approved):

src/core/user/UserService.ts
```diff
@@ -75,7 +75,6 @@ export default class UserService extends ApiClientBase implements IUserService {
   public async logout() {
     this.hasUser = false;
```
andyguest, line 76: Why don't we just remove this.patientData here too?   And any other things that may now be invalid.

Or if we can't do that, see my other question.

src/features/AppCoordinator.ts
```diff
@@ -68,6 +68,8 @@ export class AppCoordinator extends Coordinator implements SelectProfile, Editab
 
   shouldShowCountryPicker: boolean = false;
 
+  thisUser: UserResponse | null;
+
   screenFlow: Partial<ScreenFlow> = {
     Splash: () => {
       if (this.patientData && this.shouldShowCountryPicker) {
```
andyguest, line 75: It's not safe for us to look at this.patientData before validating we have this.thisUser.  I think this sequence of code arms should be reordered / protected better.  

Is there potential for this kind of bug to exhibit when we remove and reinstall?  Did we make the necessary changes to remove ALL local data on uninstall of our app?

andyguest (commented):
I've attached a couple of questions to this.  They don't need to impact this PR, but might be food for further thought.

src/features/AppCoordinator.ts
selvach: AsyncStorage is wiped on uninstall on iOS and Android.

PR #986, opened 2020-12-08
Fix incorrect home screen when US users goes back from thank you screen

the-kenneth:
Main fix is in file assessmentCoordinator line 202 and 208.

The rest is a rename of the thank you screens to reflect their usage.


selvach:
One impact of the re-naming of the screens is it impacts Amplitude analytics. I think I'm okay with this tradeoff.... we'll have to adjust the charts we have saved - but that's not a big deal. WDYT?

selvach (approved):

the-kenneth:
> One impact of the re-naming of the screens is it impacts Amplitude analytics. I think I'm okay with this tradeoff.... we'll have to adjust the charts we have saved - but that's not a big deal. WDYT?

I think correcting this in amplitude will also help there.

PR #988, opened 2020-12-09
Resets stack after finishing daily dose symptoms report

the-kenneth:
Stops users going back to report dose symptoms.


selvach:
NICE! Can you do this on the VaccineThankYouScreen as well?

selvach (changes requested):
NICE Can you do this on the ThankYouScreen as well, please!

selvach (approved):

PR #992, opened 2020-12-10
Vaccine 1.1 changes

the-kenneth:
Removed 2 placebo/trial screens.
Updated copy. 
Send vaccine on submit of symptoms button if screen opened from new vaccine flow.


src/features/vaccines/VaccineDoseSymptomsScreen.tsx
```diff
@@ -27,18 +28,26 @@ export const VaccineDoseSymptomsScreen: React.FC<Props> = ({ route, navigation }
...
+        //TODO Show error message toast?
+      } finally {
+        assessmentCoordinator.gotoNextScreen(route.name);
+      }
+    }
   };
 
   const registerSchema = Yup.object().shape({}).concat(DoesSymptomsQuestions.schema());
-  const currentPatient = assessmentCoordinator.assessmentData.patientData.patientState;
+  const currentPatient = route.params.assessmentData.patientData.patientState;
```
selvach, line 50: Thanks for cleaning this up. Just for my understanding - was there something "wrong" with this? As assessmentCoordinator is a singleton, right? I can see how this might make testing a bit easier - as you're injecting route.params to the screen.

selvach (approved):

src/features/vaccines/VaccineDoseSymptomsScreen.tsx
the-kenneth: In an ideal code base I would have the assesstmentData private and encapsulated ( and screens have their coordinator passed in via the props as well.. but thats something else). A lot of the old assessment screens access it directly but i would like to clean that up at some point.

PR #996, opened 2020-12-10
OTA Release to 2.1.0

selvach:
Feature
 - New question on Vaccines!
 - New Blog on Asymptomatic COVID, Tim's weekly video and recording on the Vaccine Rollout Seminar -

Bug Fixes:
 - Fixes incorrect Homescreen when users return from the Thank You Page
 - Fixes wrapping of text on slim phones on the Profile Selection page
 - Typo / Copy Change for Swedish app.


src/components/SelectorButton.tsx
```diff
@@ -1,22 +1,26 @@
...
 
 import { colors, fontStyles } from '@theme';
 
-import { BigButton, IButtonPress } from './BigButton';
 import { FieldWrapper } from './Screen';
 
-interface SelectorButtonProps {
+type IButtonPress = ((event: GestureResponderEvent) => void) | undefined;
+
+interface Props {
```
petermason, line 12: suggestion. name props IProps to identify it's an interface

src/core/vaccine/dto/VaccineRequest.ts
```diff
@@ -0,0 +1,62 @@
+export type VaccineRequest = {
```
petermason, line 1: like interface names start with a capital 'I' its a good convention to identify types with a capital 'T'. This hasn't been widely use in this project but thought I'd make a comment.

the-kenneth (approved):

src/CovidApp.tsx
```diff
@@ -223,8 +230,8 @@ export default class CovidApp extends Component<object, State> {
         <Stack.Screen name="WhereAreYou" component={WhereAreYouScreen} options={noHeader} />
         <Stack.Screen name="TreatmentSelection" component={TreatmentSelectionScreen} options={noHeader} />
         <Stack.Screen name="TreatmentOther" component={TreatmentOtherScreen} options={noHeader} />
-        <Stack.Screen name="ThankYou" component={ThankYouScreen} options={noHeader} />
-        <Stack.Screen name="ViralThankYou" component={ViralThankYouScreen} options={noHeader} />
+        <Stack.Screen name="ThankYouSE" component={ThankYouSEScreen} options={noHeader} />
```
petermason, line 233: Note: future updates to Navigation need to be made in src/routes/main-navigator

selvach: Do they need to be made in both right now?

src/components/SelectorButton.tsx
selvach: Nice! I like this suggestion - lets pick these up in PRs into develop! Maybe share in the slack channel or start on Notion Page on code styles? You could also include the changes you do to introduce index.ts files, etc.

src/CovidApp.tsx
selvach: You mentioned the app would crash without it?  What if one screen is missing from one of the lists?

petermason: main-navigator is the correct place to put the new routes. The references are only in CovidApp because there is a strange issue when they are remove. However they are no longer used.

PR #1009, opened 2020-12-15
Vaccine v2

the-kenneth:
All ready for testing now. A number of V1 screens are no longer used, but have yet to be removed.


selvach:
@the-kenneth @snack-0verflow I've made the following changes:

## Logic for showing the VaccineList 
When navigating from CovidTestList, we have to check `shouldShowVaccineList` to determine whether we should the VaccineList screen or start the DailyAssessment (HowYouFeel). This flag is set from the backend on the getPatient() call. The logic in the backend will controll rollout to country (currently UK only, will be changed to include US later). It also checks the age and doesn't show for children. 

## Logic for navigating to the next step after VaccineLIst
- When navigating from the VaccineList to the next screen, the logic is now:

-- If the user has no Vaccines (check the length of vaccines response) and has not filled in a VaccinePlan (check the server) - then navigate to VaccineHesitancy screen

-- If the user has no Vaccines (check the length of vaccines response) and has already filled in a VaccinePlan THEN navigate to the DailyAssessment flow. 

-- If the user has vaccines entered - then iterate through the all the doses and see if the date is within the valid range. (i.e. date is within the last week). This check used to be done on the server in the v1 flow. I think it makes more sense to do it in the UI, seeing as we have all the information. 

## Tidy Up
- Remove all the v1 screens
- Remove the `recordVaccine` logic on the DoseSymptoms screen
- One small typo

selvach (dismissed):

src/core/assessment/AssessmentCoordinator.ts
```diff
@@ -155,6 +128,20 @@ export class AssessmentCoordinator extends Coordinator {
     OtherSymptoms: () => {
       NavigatorService.navigate('WhereAreYou', { assessmentData: this.assessmentData });
     },
+    AboutYourVaccine: () => {
+      NavigatorService.goBack();
+    },
+    VaccineList: (params: { shouldAskDoseSymptoms: boolean; askVaccineHesitancy: boolean; dose: string }) => {
```
the-kenneth, line 134: I think the type should be `boolean | undefined`

selvach: Thanks for the spot. Updated!

selvach (approved):

selvach:
@the-kenneth As you raised the PR, I can't add you as an approver. I've approved. If you're happy - could you merge to develop, please?

selvach (approved):

PR #1025, opened 2020-12-23
Remove donate module

selvach:
Remove unused code


src/components/Donate.tsx
```diff
@@ -1,46 +0,0 @@
-import React, { Component } from 'react';
```
andyguest, line 1: (Surprised not to see any reference to Donate.tsx being imported; do we import * from src/components presumably?)

andyguest (approved):
Looks fine.

src/components/Donate.tsx
selvach: It was imported directly onto the screen. But it was removed from the ThankYou Page and/or Homescreen a while back.

PR #1038, opened 2020-12-31
Remove console logs

selvach:
@petermason raised that these can be a memory hog and are should not be in production code. 
I've left console logs in the `compare.ts` which is a test script.

### DIFF
--- a/src/core/api/ApiServiceErrors.ts
+++ b/src/core/api/ApiServiceErrors.ts
@@ -72,8 +72,6 @@ export const handleServiceError = (error: ReceivedError) => {
     throw new OfflineException(error.message);
   } else if (error.code && [CONNECTION_ABORTED, TIMED_OUT].includes(error.code)) {
     throw new OfflineException(error.message);
-  } else {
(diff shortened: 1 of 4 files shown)


src/core/api/ApiServiceErrors.ts
```diff
@@ -66,15 +66,14 @@ export const handleServiceError = (error: ReceivedError) => {
...
-        console.log('[ERROR] Unhandled status:', error.response.status, ':', error.message);
         throw new ApiException(error.message, error.response.status);
     }
   } else if (error.message === NETWORK_ERROR) {
     throw new OfflineException(error.message);
   } else if (error.code && [CONNECTION_ABORTED, TIMED_OUT].includes(error.code)) {
     throw new OfflineException(error.message);
   } else {
-    console.log('[ERROR] Unhandled error:', error.message);
+    // console.log('[ERROR] Unhandled error:', error.message);
```
petermason, line 76: dead code, maybe throw an error?

the-kenneth (approved):

PR #1048, opened 2021-01-06
Feature/track diet score playback

petermason:
# Description

Add an app redux state to track global app states. Add first element to track 'dasboardVisited'

use dashboardVisited to send analytics event once for Diet Study Card (if displayed)

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


the-kenneth (changes requested):
This wouldn't scale to analytics on other cards on the dashboard, as the visited state would already be true. Wdyt?

petermason:
> This wouldn't scale to analytics on other cards on the dashboard, as the visited state would already be true. Wdyt?

This is designed to prevent multiple tracking events as the dashboard can be loaded multiple times through the use of the app. So here is used to basically say the dashboard has loaded. The reducer can be expanded for future app states that need to be tracked.

the-kenneth (approved):

PR #1054, opened 2021-01-07
add 3 x scale images for icons where svg cannot be used

petermason:
# Description

add 3 x scale images for icons where svg cannot be used

Note: Ideally these should be svg not png

Use images 3 x required size to allow for pixel density

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots
image

## Checklist

- [ ] I have run `npm test`


selvach:
@petermason Would this approach here be of use:
https://docs.expo.io/versions/v36.0.0/react-native/images/

petermason:
> @petermason Would this approach here be of use:
> https://docs.expo.io/versions/v36.0.0/react-native/images/

I'm simply loading the larger version and setting the size to an absolute amount so will work for all devices.

selvach (approved):

PR #1061, opened 2021-01-11
Max 500 chars in other input

the-kenneth:


selvach:
Hmm.. this doesn't appear to fix anything? 

The behaviour I see in iOS simulator when typing more the 500 characters into this box on Dose Symptoms is that it throws a 400 Validation Error on the backend (>500 characters). The UI swallows this error and proceeds to the next screen.... 

Screenshot 2021-01-11 at 14 11 17

the-kenneth:
Ive now also added it to the dose symptoms field. It was only on the hesitancy one

bionara:
Code looks good. Have pulled latest master in.

bionara (approved):

PR #1068, opened 2021-01-13
Tidy school networks redux and remove dead screen

the-kenneth:
- Removed unused variables from school slice, added delete reducer and tidied existing reducer.
- Fixed bug where local state was out of sync with server when deleting school by also removing from local state on success.
- Renamed some variables and functions to make them clearer

_Currently our redux state files have different files/formats and these should be aligned at some point_


selvach (approved):
Nice clean up!

PR #1069, opened 2021-01-13
Remove "thank you" page after vaccination & replace with message/more info screen.

bionara:
# Description

**Remove vaccine "thank you" page and add conditional message to the "How do you feel" question page.**

https://www.notion.so/joinzoe/Remove-vaccine-thank-you-page-and-add-conditional-message-to-How-do-feel-question-b6881ac1724b42ffa9bbe0e559700a8f

PR maybe looks a bit chunky for two reasons:
1. We moved the fetch vaccines call/code to redux
2. Addition of new page (the info modal) with associated routing setup and language files.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

image

(blue text links to:)
image

## Checklist

- [x] I have run `npm test`

N/A


assets/lang/en-US.json
```diff
@@ -63,6 +63,12 @@
   "vaccines": {
     "vaccine-list": {
       "description": "Tell us if you’ve had at least one dose of a COVID-19 vaccine.\n\nWe can't record COVID-19 vaccines taken as part of a trial yet. If this is how you received your vaccination, please don't add it for now. "
+    },
+    "log-symptoms": {
```
the-kenneth, line 67: If the en-US text is the same as the base en text then the keys are not needed and it defaults to the en file. This is useful because then it means the en-US file only has US changes. So I think these can be removed.

assets/lang/es.json
```diff
@@ -1076,6 +1076,13 @@
         "next": "Next"
```
the-kenneth, line 1076: Missing keys from the HowYouFeel screen:

"update-location": "",
"vaccine-reporting-message": "",
"vaccine-reporting-link": ""

assets/lang/sv-SE.json
```diff
@@ -1031,6 +1031,12 @@
         "not-to-say": "",
```
the-kenneth, line 1031: Also missing keys from the HowYouFeel screen:

"update-location": "",
"vaccine-reporting-message": "",
"vaccine-reporting-link": ""

src/core/state/vaccines/slice/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    const patientVaccines = response.filter((vaccine) => vaccine.patient === patientId);
+    return patientVaccines;
+  }
+);
+
+const vaccinesSlice = createSlice({
+  name: 'Vaccine',
+  initialState,
+  reducers: {
+    setVaccine: (state, action: PayloadAction<VaccineRequest[]>) => {
```
the-kenneth, line 28: In the end this was not used, maybe we can comment it just to mention its currently unused or wdyt?

the-kenneth (changes requested):
Really nice work Ben! Touched on lots of essential ideas and concepts we use in our code base and made this work nicely. Just 1 or 2 missing keys but thats it.

assets/lang/en-US.json
bionara: Removed the block

assets/lang/es.json
bionara: Added in

src/core/state/vaccines/slice/index.ts
bionara: Added a comment - it may be useful for the stuff we talked about in standup today regarding POST/saving

assets/lang/sv-SE.json
bionara: Added

the-kenneth (approved):

PR #1072, opened 2021-01-18
Fix nav stack bug and remove dead code

the-kenneth:
- Resets nav stack after vaccine hesitancy question.
- Removes dead VaccineThankYou code


selvach (changes requested):
Could you kill any unused i18n strings as well, please?! 

Otherwise: looks good!

selvach (approved):

PR #1079, opened 2021-01-22
v3 vaccine edit

bionara:
# Description

Rejigs the add vaccine screen(s) as per the new design to gather further data.
 Full requirements here: 
https://www.notion.so/joinzoe/017c3de591604a7b89d0b69e3c37b279?v=0133357df3ed42e1bd42c1e0d7afcf45&p=8c3e4d43cdfe459da2d65f5426b1b94d
Design: https://www.figma.com/file/7TEcX7f6rw3Z0WYR4TpX3R/COVID-Symptom-Study-(App)---MASTER?node-id=6743%3A3918&viewport=-2281%2C354%2C0.5

## How to test locally
Run the corresponding BE branch if not merged https://github.com/zoe/covid-app/pull/387 . If merged, point against staging or use master.

Access page: home > profile > not had test.
Observe no vaccines + "Add Vaccine" button
Tap "Add Vaccine" button. US/GB users should have a blue link to another screen with image examples (SE not).
Add a DOSE ONE vaccine. To start, just add a date and save. This will let you see the incomplete vaccine/doses list, plus the prompt edit.
From here, edit vaccine and complete fields as required. Note changing the date in edit mode will prompt alert on save.
Delete vaccine now at bottom of dose in edit mode only.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Vaccine with a dose that is missing a date (easiest way to replicate is to create a dose, save it, then manually delete its date in django admin)
image

Dose 1 complete
image

Both doses complete
image

Example vaccine doc screen
image

## Checklist

- [ ] I have run `npm test`

Approach followed the Formik form model in which one parent form is passed down to the props, and reused props check if first dose and select properties based on that. Would be good to make this less rigid.
Setting second dose to 'no' on edit mode doesn't delete it yet. Ticket created.


src/features/vaccines/AboutYourVaccineScreen.tsx
```diff
@@ -26,51 +30,66 @@ type Props = {
...
-interface AboutYourVaccineData extends VaccineDateData {}
+const registerSchema = Yup.object().shape({}).concat(VaccineDoseQuestion.schema());
+
+interface AboutYourVaccineData extends VaccineDoseData {}
 
 export const AboutYourVaccineScreen: React.FC<Props> = ({ route, navigation }) => {
   const vaccineService = useInjection<IVaccineService>(Services.Vaccine);
   const coordinator = assessmentCoordinator;
-  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [errorMessage] = useState<string>('');
```
petermason, line 40: no setter and the default value is an empty string?

assets/index.ts
```diff
@@ -100,6 +100,10 @@ export const dietStudyPlaybackReady = require('./diet-playback-ready.png');
 // share svg
 export { default as ShareIcon } from './icons/Share';
 
+// Vaccination example sheets
+export const vaccinationExampleUK = require('./vaccines/vaccine-demo-uk.png');
```
petermason, line 104: can you check the size of these. Small icons render poorly on higher pixel density devices. Convert to SVG?

src/features/vaccines/AboutYourVaccineScreen.tsx
bionara: Removed as it's unused - global error message at bottom instead.

assets/index.ts
bionara: good shout! assumed they were png in the image, but are  svg and can be exported. Doing now.

bionara: Updated those to SVGs

src/components/Screen.tsx
```diff
@@ -117,7 +117,11 @@ export default class Screen extends Component<ScreenProps> {
                 flexGrow: 1,
                 justifyContent: 'space-between',
               }}>
-              <View style={styles.pageBlock}>{this.props.children}</View>
+              {this.props.extendEdges ? (
```
bionara, line 120: This prop was available on screen, but not used at all in the codebase. I am using it on the vaccine info screen, to set <Screen> margin to 0, then wrap some content (but not the full width image), whilst keeping everything within the screen for scrolling.

petermason: makes sense

petermason (approved):

selvach:
This is a great peace of work @bionara !

PR #1080, opened 2021-01-25
Fix mis-matched translation string

selvach:
Fixes this mismatched i18n strings:

Screenshot 2021-01-25 at 18 06 55

Screenshot 2021-01-25 at 18 41 21


bionara (approved):
Looks good!

PR #1081, opened 2021-01-25
Rename US Cohort

selvach:
- This PR renames the display name of a US Study Cohort, as requested by Andy Chan. 
- Fixes padding that was a bit of an eye sore

## Before:

Screenshot 2021-01-25 at 18 49 16

## After 

Screenshot 2021-01-25 at 18 48 49

### DIFF
--- a/src/features/patient/YourStudyScreen.tsx
+++ b/src/features/patient/YourStudyScreen.tsx
@@ -342,7 +342,7 @@ export default class YourStudyScreen extends Component<YourStudyProps, State> {
               <Form>
                 <FieldWrapper>
                   <Item stackedLabel style={styles.textItemStyle}>
-                    <Label>{i18n.t('your-study.label-cohort')}</Label>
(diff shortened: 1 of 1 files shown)


bionara:
Nice work on the padding!

src/features/patient/YourStudyScreen.tsx
```diff
@@ -101,7 +101,7 @@ const AllCohorts: CohortDefinition[] = [
   },
   {
     key: 'is_in_us_origins',
-    label: 'ORIGINS',
+    label: 'ORIGINS-Columbia University',
```
bionara, line 104: Is this space not an issue? 
(general Q, no reason to think it may be, just a bit odd to have spaces in labels)

bionara (approved):

src/features/patient/YourStudyScreen.tsx
selvach: This was the exact text that Andy suggested

PR #1088, opened 2021-01-28
Bug fixes on vaccine form date and button

bionara:
# Description

_Description of the feature or fix with link to the issue if existing_

## On what platform have you tested the change?

- [x] Android - Emulator
- [ ] Android - Physical device
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

_Please attach screenshots showing the change if relevant_

## Checklist

- [ ] I have updated mockServer
- [ ] I've added analytics as relevent
- [ ] I have run `npm test`
- [ ] I have run `npm run test:i18n`

## Out of scope and potential follow-up

_Is there any related changes that you plan to do in a follow-up PR or voluntarily excluded from the scope?_

PR #1089, opened 2021-01-28
Make lot number optional

selvach:
Currently the formik validation is incorrectly blocking me from editing a vaccine (e.g. to set the type) and leaving the Batch Number / Lot Number blank. 

Screenshot 2021-01-28 at 19 38 54

Quick fix for this!


bionara (approved):
Wierd, i had done this on my PR (https://github.com/zoe/covid-tracker-react-native/pull/1088) but must've reverted it before committing sorry. Merging.

PR #1106, opened 2021-02-02
Vaccine hesitancy seen manager for screen

bionara:
# Description

**Requires BE ticket and migration to work!** https://github.com/zoe/covid-app/pull/391

Adds a new boolean field to the Patient model
Uses new field to determine link route from vaccine list page to hesitancy page
Updates new field to True on the first time it is triggered (ie user redirected and sees page)

Test:
Go to "Your covid 19 vaccines" page > delete any vaccines > tap "I haven't had a vaccine" > observe redirect to hesitancy page
hit back (state updates on patch above)  > tab "I haven't had a vaccine" > observe redirect to "How do you feel" page

You can reset the field on django admin at /admin/users/patient/, right at page bottom.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Regarding this button "I haven't had vaccine" and its redirect.
image

## Checklist

- [ ] I have run `npm test`


bionara:
Closing as this is a none bug due to the check being based on Vaccine Plan lookup, and not Vaccine Doses. Test case in the ticket apparently didn't SAVE the vaccine plan, and so the UX is indeed as it should be, and this new property and PR is not needed. Less noise, till works, all happy!

PR #1112, opened 2021-02-04
Change share vaccine banner svg for png 

bionara:
# Description

Swaps the svg (has some weird margins within the svg that are not easily removable) for a png, just like the others banners.
Deletes the standalone component for the button and moves the shareImage code to the dashboard (duplicated for each - would be good to create a dashboard base class when time permits :) )

- [ ] Android - Emulator
- [x ] iOS - Emulator - IPHONE 8 and IPHONE 11
- [ ] iOS - Physical device

## Screenshots

11:
image

8
image

## Checklist

- [ ] I have run `npm test`


src/features/dashboard/DashboardScreen.tsx
```diff
@@ -100,9 +119,23 @@ export const DashboardScreen: React.FC<Props> = ({ navigation, route }) => {
       navigation={navigation}
       compactHeader={<CompactHeader reportOnPress={onReport} />}
       expandedHeader={<Header reportOnPress={onReport} />}>
-      <DashboardLogVaccine screenName="Dashboard" />
-
       <View style={styles.calloutContainer}>
+        <ExternalCallout
+          calloutID="sharev3"
```
selvach: FYI - I added a commit to change the calloutID, which is used in amplitude

src/features/dashboard/DashboardScreen.tsx
```diff
@@ -100,9 +119,23 @@ export const DashboardScreen: React.FC<Props> = ({ navigation, route }) => {
...
       expandedHeader={<Header reportOnPress={onReport} />}>
-      <DashboardLogVaccine screenName="Dashboard" />
-
       <View style={styles.calloutContainer}>
+        <ExternalCallout
+          calloutID="shareVaccine"
+          imageSource={shareVaccineBanner}
+          aspectRatio={311 / 135}
+          screenName={route.name}
+          postClicked={() =>
```
selvach, line 128: Ah! You don't need the postClicked here. The `ExternalCallout` component does analytics for you!

selvach (approved):

PR #1118, opened 2021-02-04
Remove random files

selvach:


bionara:
Cheers - saw these go in git and assumed they were "normal" so left in.
just had a peep now https://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them 
Will remove any in future!

PR #1128, opened 2021-02-08
V3Vaccine force user to update missing data

bionara:
# Description

Disables the vaccine list CTA when there is missing required data on dose 1 or 2 - date, brand and description.

image

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


bionara:
Note that after a chat with Philipp about design, any missing field prompts are now all in orange.
When in grey, as per old design, it was difficult to see what was needed and why the CTA was disabled (it kind of still is, but the orange helps).

image

selvach (approved):
Pulled down and tested! Looks great.

PR #1129, opened 2021-02-08
Type checks on Diet Scores

selvach:
- Add type checks on diet scores.
- Add analytics events when someone removes themselves from the mailing list


bionara (approved):
looks good!

PR #1130, opened 2021-02-08
OTA Release 2.4.0

selvach:
Fixes:
 - Removes popup about date change when editing vaccines
 - Enforces every vaccine logger to add they Vaccine Type


bionara (approved):
Checked out & ran in emulator. Tested the vaccine edit & validation. Approved for OTA via master merge.

PR #1131, opened 2021-02-08
Change tap target for editing vaccine

selvach:


bionara (approved):
Code looks good. Checked out and ran on emulator, works fine. nice one!

PR #1132, opened 2021-02-08
Bug Fix for Diet Study Feedback Crashes

selvach:
The PR fixes a bug where are user can enter the DietStudyPlayback flow, while this.patientData is set to secondary profile. 

1) Only should the DietStudy Invite on the ThankYou page, if reporting for the primary profile. 
2) When starting the flow (which can happen from the Dashboard) with this.patientData set to secondary profiles force a reset. (This currently incurs an extra API call, which ideally should not be necessary - but requires a bigger plumbing and moving patientLists and primary patientID into redux). 

- This kind of glosses over the question of "What should this.patientData point to, when you are on the Dashboard, after just finishing with reporting for a secondary profile. This is MUCH bigger re-architecture.

### DIFF
--- a/src/features/thank-you/ThankYouUKScreen.tsx
+++ b/src/features/thank-you/ThankYouUKScreen.tsx
@@ -55,7 +55,11 @@ export default class ThankYouUKScreen extends Component<RenderProps, State> {
   async componentDidMount() {
     const { startupInfo } = store.getState().content;
     const variant = startExperiment(experiments.UK_DietScore_Invite, 2);
-    const showDietStudyPlayback = (variant === 'variant_2' && startupInfo?.show_diet_score) || false;
+    const showDietStudyPlayback =
+      (variant === 'variant_2' &&
+        startupInfo?.show_diet_score &&
+        !appCoordinator.patientData.patientState.isReportedByAnother) ||
+      false;
 
     if (showDietStudyPlayback) {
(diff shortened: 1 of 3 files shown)


src/features/thank-you/ThankYouUKScreen.tsx
```diff
@@ -55,7 +55,11 @@ export default class ThankYouUKScreen extends Component<RenderProps, State> {
   async componentDidMount() {
     const { startupInfo } = store.getState().content;
     const variant = startExperiment(experiments.UK_DietScore_Invite, 2);
-    const showDietStudyPlayback = (variant === 'variant_2' && startupInfo?.show_diet_score) || false;
+    const showDietStudyPlayback =
+      (variant === 'variant_2' &&
```
bionara, line 59: Move 'variant_2' to a const and name it appropriately.

bionara (approved):
Approved.

src/features/thank-you/ThankYouUKScreen.tsx
selvach: So... I dynamically set this based on this code:
https://github.com/zoe/covid-tracker-react-native/blob/c26d1fff67ebb8a07b1e34d66474c53e4d4bba0c/src/core/Experiments.ts#L24

It doesn't lead to the most elegant code....

PR #1135, opened 2021-02-10
Filter the universities on the backend

andyguest:
The plain getSchools() API call is only used for the purposes of listing universities (of which we have very few).
In all other use cases we have additional filtering we can apply (select by school id or member_verify_code).

# Description

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


bionara (approved):
Code looks good!

andyguest:
No great value in pursuing this - largely an unused feature.

PR #1162, opened 2021-02-23
Refactor: Remove mental health intro screen

selvach:
Removes the dead screen and code


bionara (approved):
hope it doesn't get reinstated in 3 days!

PR #1167, opened 2021-02-23
Wrap Mental Health Modal in scroll view

selvach:
- For the video, I created some fake padding in order to simulate a model that goes off the screen. 
- All this PR does is wrap the modal in scroll view. 
- Also fixes Karestan's name

Screen Recording 2021-02-23 at 20 36 03


bionara (approved):
good solution. We should probably look into something that uses/mimics the SafeArea

PR #1175, opened 2021-03-01
Update App modal and logic

bionara:
# Description

Set as Draft until we can prettify the modal (and address any UX concerns).

Mechanism to show a "please update" modal on the app when we deem the user should.
This is based on a Responsible Developer code, in which we only update the API middle version number when we introduce a breaking change. The table below shows this:

```
BE      FE       Show modal?
2.4.0   2.5.0    No
2.4.0   2.4.3    No
2.4.0   2.4.0    No
2.4.0   2.3.99   Yes
2.4.0   1.3.0    Yes
```

Note that the modal won't be shown if the app's version gets ahead of the API.

**How to test:**
Locally, this is very easy:  go into the django settings and set `API_VERSION` to `9.9.9`. Refresh your app ("r" on the console). Modal should appear. 
If using staging, you can hardcode the version to trigger and test the UI: find this method and make it return`true` `checkVersionOfAPIAndApp`

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

ugly screenshot. I am in talks with Christina and awaiting a bit of direction from the new designer.
image

Note that this shouldn't have a close/cancel/cta as it should block the user app access.
note two: modal probably wants some info on _how_ to update.

## Checklist

- [ ] I have run `npm test`


src/core/content/ContentService.ts
```diff
@@ -70,9 +72,35 @@ export default class ContentService implements IContentService {
...
+  async checkVersionOfAPIAndApp(apiVersion: string | undefined): Promise<boolean> {
+    const appVersion: string | undefined = Constants.manifest.version;
+    if (!appVersion || !apiVersion) {
+      // Error getting versions - better to return false than incorrectly show a "please update" popup
+      return false;
+    }
+
+    /*
+      We are actually only concerned about the "middle" version number
+      as our Best Practices at versioning means an increment of the "middle number"
```
andyguest, line 84: I think I would say that this sounds ever-so-slightly fragile and you've done the splitting already, so why not include any advance on the first number too, to allow a 3.0 -> 4.0 upgrade in the future if we ever wished to do so.  Otherwise we sort of have to remember not to ever do that, I think...

src/core/VersionUpdateModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  route: RouteProp<ScreenParamList, 'VersionUpdateModal'>;
+};
+
+export const VersionUpdateModal: React.FC<Props> = ({ navigation, route }) => {
+  const goToAppStore = () => {
+    openWebLink(getPlatformStoreLink);
+  };
+
+  useEffect(() => {
+    navigation.addListener('beforeRemove', (e) => {
```
petermason, line 24: this listener either needs to be returned or a removed eventlistener added

selvach (approved):

selvach:
I'm happy with the PR. I believe it can be rolled to production and nothing will happen. (i.e. no one  will see the modal, until we upgrade to 2.5.0 via the backend flag). 

FYI - I've made the following changes:
 - I've fixed a merge conflict https://github.com/zoe/covid-tracker-react-native/pull/1175/commits/e6c4aa86a9deeda7c487005c32a8114b49f39018
 - Ensure we clean up properly after the useEffect - https://github.com/zoe/covid-tracker-react-native/pull/1175/commits/39e278f02729cb091d17c90fbd4ee9f0d7157c71
 - Refactored for style https://github.com/zoe/covid-tracker-react-native/pull/1175/commits/94283dda8c99cf9a4a4d2b3b3d61c3119727f818


@petermason @bionara - If you are happy also, please hit "Squash and Merge"!! Thanks!

PR #1026, opened 2020-12-23
prevent multi tap: login, register

petermason:
# Description

Prevent the multi or unnecessary tap of the submit button on forms

## On what platform have you tested the change?

- [x] Android - Emulator
- [ ] Android - Physical device
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

# LOGIN:

### idle state: 
image

### in progress:
image

### ready
image

### post submit with error
image

# REGISTER

### idle state
image

### in progress:
image

### ready
image

### post submit with error
image

# TEST

### Submitting

image

_Please attach screenshots showing the change if relevant_

## Checklist

- [ ] I have updated mockServer
- [ ] I've added analytics as relevent
- [ ] I have run `npm test`
- [ ] I have run `npm run test:i18n`

## Out of scope and potential follow-up

_Is there any related changes that you plan to do in a follow-up PR or voluntarily excluded from the scope?_

PR #1179, opened 2021-03-02
add is submitting logic to pages without any

petermason:
# Description

Prevent multiple clicks during transion between screens

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/features/assessment/HowYouFeelScreen.tsx
+++ b/src/features/assessment/HowYouFeelScreen.tsx
@@ -1,6 +1,6 @@
-import { RouteProp } from '@react-navigation/native';
-import { StackNavigationProp } from '@react-navigation/stack';
 import React, { useEffect, useState } from 'react';
+import { RouteProp, useIsFocused } from '@react-navigation/native';
+import { StackNavigationProp } from '@react-navigation/stack';
 import { TouchableOpacity } from 'react-native';
 import { View } from 'native-base';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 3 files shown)


andyguest (commented):
Peter, I was looking back to an earlier fix on the RegisterScreen - my evidence in the server logs suggests this fix may not actually be enough to prevent the double tap.  Can you do some local testing perhaps firing off toasts instead of sending reports to the backend to see if double hits remain possible?  

(It may be that I haven't identified the true source of the double taps though.)

https://github.com/zoe/covid-tracker-react-native/pull/1026/files

petermason:
> Peter, I was looking back to an earlier fix on the RegisterScreen - my evidence in the server logs suggests this fix may not actually be enough to prevent the double tap. Can you do some local testing perhaps firing off toasts instead of sending reports to the backend to see if double hits remain possible?
> 
> (It may be that I haven't identified the true source of the double taps though.)
> 
> https://github.com/zoe/covid-tracker-react-native/pull/1026/files

I'm going to take another look at register and make a separate pr if required. This PR the logic is simpler, preventing additional taps after the first. This is achieved by changing a state variable which forces react to rerender the view, additional taps are blocked during this cycle. Once this cycle is complete the isSubmitting has been changed to true, thereofore, preventing additonal submissions. This state is only changed if the user returns to the page (back) or an error in the submission occurs.

bionara (approved):
Good looks good and is a logical fix to the double submit!

src/features/assessment/HowYouFeelScreen.tsx
```diff
@@ -93,6 +100,10 @@ export const HowYouFeelScreen: React.FC<Props> = ({ route, navigation }) => {
 
   const currentPatient = assessmentCoordinator.assessmentData.patientData.patientState;
 
+  useEffect(() => {
```
selvach, line 103: @petermason Sorry I didn't get to review this at the time. @petermason Could you explain what this does, please?

PR #1199, opened 2021-03-13
Add Johnson Vaccine for US

selvach:
- This PR separates out the vaccines brand options between UK, SE and US instead of using splice for clarity
- Add's the J&J vaccine option for the US
- Hides the second dose on the Vaccine Card (List Display) and on the "AboutYourVaccine" screen (Create / Edit Vaccine)

This PR works with the fact that all the vaccine code was written with the assumption of 2 doses, rather than try to addrss that assumption.  When we get another vaccine with either 1 dose or 3 doses then we should refactor. For now and "if vaccine.brand == johnson"

### DIFF
--- a/src/core/vaccine/dto/VaccineRequest.ts
+++ b/src/core/vaccine/dto/VaccineRequest.ts
@@ -1,3 +1,5 @@
+import i18n from '@covid/locale/i18n';
+
 export type VaccineRequest = {
   id: string;
   version: string; // document/schema version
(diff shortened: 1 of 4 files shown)


src/core/vaccine/dto/VaccineRequest.ts
```diff
@@ -1,3 +1,5 @@
+import i18n from '@covid/locale/i18n';
+
 export type VaccineRequest = {
```
petermason, line 3: like we prefix interfaces with an 'I' can I suggest we use 'T' for types

selvach: Happy to adopt that! Seeing as these types were already here before this PR - can I open another PR to refactor the names?

petermason: 👍

petermason (approved):

PR #1209, opened 2021-03-19
Selvach/refactor buttons

selvach:
- Moves BrandedButton and BigButton into the buttons subdirectory
- Rename button directories into lowercase hypen-case
- Sort out the rubbish import tree. 
- BrandedButton was imported into a file called Text and the imported everywhere all across the code base. This was just get crazy...

In a future PR:
 - Replace Button, ClearButton and ActionButton


bionara (approved):
Looks good, nice clean up!

petermason (approved):

PR #1213, opened 2021-03-25
Set user signup date from profile for timeline modal

bionara:
# Description
Updates the timeline modal's date placeholder to use the date from the current user - assumed that this is the core user as it happens on app start

kind of confusing that the screenshot says feb... 3,2,1... design critique incoming ! :)

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

image

## Checklist

- [ ] I have run `npm test`

no

### DIFF
--- a/src/features/anniversary/screens/anniversary-modal.tsx
+++ b/src/features/anniversary/screens/anniversary-modal.tsx
@@ -1,17 +1,35 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Image, ScrollView, StyleSheet, TouchableOpacity, View } from 'react-native';
 import { useNavigation } from '@react-navigation/native';
 import { useDispatch } from 'react-redux';
+import moment from 'moment';
 
 import { SafeLayout, Text } from '@covid/components';
 import { timelineModalCard } from '@assets';
 import { setHasViewedAnniversaryModal } from '@covid/core/state';
+import { useInjection } from '@covid/provider/services.hooks';
+import { IPatientService } from '@covid/core/patient/PatientService';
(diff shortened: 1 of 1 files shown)


src/features/anniversary/screens/anniversary-modal.tsx
```diff
@@ -1,17 +1,35 @@
...
+import { Services } from '@covid/provider/services.types';
+import { Profile } from '@covid/components/Collections/ProfileList';
 
 import appCoordinator from '../../AppCoordinator';
 
 function AnniversaryModal() {
   const { goBack } = useNavigation();
   const dispatch = useDispatch();
+  const patientService = useInjection<IPatientService>(Services.Patient);
+  const [signupDate, setSignupDate] = useState<string>('');
```
andyguest, line 21: is it not possible just to pass this date in as a property for the component instead of digging the value out from a service?

bionara: Can do, but parent (DashboardScreen) doesn't have that property, so would either need to request it via the service, or have it passed in. 
Leaving the request in the timeline modal mean that it's easy to cleanup when we remove the modal code, and also that dashboard doesn't have redundant code for when the modal isn't called (ie ha been seen).

andyguest: Does the async useEffect mean that we get a FOUC-like effect as the View re-renders once the date answer arrives?  Or are there react rules that mean the first render will have the answer?

andyguest (approved):
OK then :-)

PR #1215, opened 2021-03-28
PCR Test Changes & Swedish Test Changes

selvach:
This PR does the following:

- Updates the copy to include instructions on lateral flow tests. 
- Always shows the lateral flow / is this a PCR test yes/no question. (It was previously conditionally displayed, which confused users)
- Updates Swedish copy (more general changes)
- Add's a Swedish only drop down option.


src/core/user/dto/UserAPIContracts.ts
```diff
@@ -39,6 +39,7 @@ export enum CovidTestMechanismOptions {
   NOSE_SWAB = 'nose_swab', // Deprecated
   THROAT_SWAB = 'throat_swab', // Deprecated
   NOSE_OR_THROAT_SWAB = 'nose_throat_swab',
+  NOSE_OR_THROAT_SWAB_AND_SALIVA = 'nose_throat_swab_and_saliva', // SE Only
```
bionara, line 42: maybe good to have SE_ONLY_ as a prefix for the const so it's clear in code without comment, ie where it is used.

bionara (approved):

PR #1216, opened 2021-03-28
Remove Mental Health Study Entry Point

selvach:
This PR:

- Removes the Mental Health Study popup. This will effectively mean there is no other way to access the mental health study. We'll keep the code in here, as we might run this again in the future. 
- Removes the UK Diet Scores card from the UK Thank You page. You'll be able to get here via the Timeline Screen.

### DIFF
--- a/src/features/dashboard/DashboardScreen.tsx
+++ b/src/features/dashboard/DashboardScreen.tsx
@@ -97,30 +97,8 @@ export function DashboardScreen({ navigation, route }: IProps) {
       }
     }
     switch (settings.currentFeature) {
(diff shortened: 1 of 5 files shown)


src/features/dashboard/DashboardScreen.tsx
```diff
@@ -97,33 +97,11 @@ export function DashboardScreen({ navigation, route }: IProps) {
       }
     }
     switch (settings.currentFeature) {
-      case 'MENTAL_HEALTH_STUDY':
```
bionara, line 100: cool 🧹  !

bionara (approved):

PR #1219, opened 2021-03-29
Check for null strings on Node in the event timeline

selvach:
- Check for null strings to prevent TextRenderered error
- Aligns card padding with the figma


andyguest (approved):
LGTM

PR #1220, opened 2021-03-29
Fixed to DietScores

selvach:
- Change DietStudy card so the whole thing is tappable f496d29
- Add DietScore card to UK Homepage 44ac494
- Fix navigation to DietScore from TimelineCard with a hack b334e52

https://user-images.githubusercontent.com/7824212/112848542-1798f780-90a0-11eb-9d47-6f68fba64ada.mov

### DIFF
--- a/src/features/anniversary/partials/finding-card.tsx
+++ b/src/features/anniversary/partials/finding-card.tsx
@@ -4,6 +4,7 @@ import { useNavigation } from '@react-navigation/native';
 
 import { Icon, Link, Text } from '@covid/components';
 import { openWebLink } from '@covid/utils/links';
+import appCoordinator from '@covid/features/AppCoordinator';
 
 import { TTimelineEvent } from '../types';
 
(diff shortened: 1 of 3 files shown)


src/features/anniversary/partials/finding-card.tsx
```diff
@@ -19,10 +20,14 @@ function FindingCard({ timelineEvent }: IProps) {
     if (external_link_text && external_link) {
       return <Link linkText={external_link_text} onPress={() => openWebLink(external_link)} />;
     }
+
     if (route_name && route_text) {
-      return <Link linkText={route_text} onPress={() => navigate(route_name)} />;
+      if (route_name === 'DietStudy') {
```
bionara, line 25: note to self: one to remember for later if adding any more internal routes.

bionara (approved):

src/features/anniversary/partials/finding-card.tsx
selvach: Yeah... it's not always straightforward :(

selvach: The appCoordinator does extra stuff to fetch the API call. This is one of the things to tidy up in our Spring Clean :)

PR #1239, opened 2021-04-01
Repeated signup 403

andyguest:
# Description

Switch to version 2 of the register endpoint to prevent stray 500 error logs!

- [ X ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


andyguest:
IMPORTANT THAT WE DON'T MERGE THIS UNTIL BACKEND FIX IS APPLIED :-)

src/features/register/RegisterScreen.tsx
```diff
@@ -83,7 +83,11 @@ class RegisterScreen extends Component<PropsType, State> {
         })
         .catch((err: AxiosError) => {
           // TODO - These error messages are misleading and we could display what the server sends back
-          if (err.response?.status === 500) {
```
bionara, line 86: we still probably want to handle a 500 right?

andyguest: that's handled generically by the else.

petermason (approved):

PR #1245, opened 2021-04-06
Remove react-native-web and update sub-dependencies

selvach:
A whole bunch of sub-dependencies are included because on react-native-web which keep hitting security advisories. 

This PR
- Removes react-native-web as a top level dependency
- Updates sub-dependencies

This solves the following security issues:

- https://github.com/zoe/covid-tracker-react-native/security/dependabot/package-lock.json/y18n/open
- And `xmlDom 0.5.0`


bionara (approved):
Good spot!

PR #1252, opened 2021-04-09
Fetch pushToken from Expo every week

andyguest:
Let's be more defensive about this, even though according to Expo themselves, push tokens should remain valid for as long as the app remains installed.  See: https://forums.expo.io/t/exponentpushtoken-when-does-it-really-change/3611/2

Let's also clear the push token if someone asks us to clear data.

N.B. UNTESTED: want to hear any comments before verifying (perhaps in Stage).

# Description

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`


src/core/push-notifications/PushNotificationService.ts
```diff
@@ -70,51 +70,48 @@ export default class PushNotificationService {
...
+    } catch (error) {
+      // Fail silently
     }
   }
 
   static async openSettings() {
     switch (Platform.OS) {
       case 'ios':
-        Linking.openURL('app-settings:');
+        await Linking.openURL('app-settings:');
```
selvach, line 111: Remove this await. It causes issues.

src/core/push-notifications/PushNotificationService.ts
```diff
@@ -70,51 +70,48 @@ export default class PushNotificationService {
...
         break;
       case 'android':
-        IntentLauncher.startActivityAsync(IntentLauncher.ACTION_NOTIFICATION_SETTINGS);
+        await IntentLauncher.startActivityAsync(IntentLauncher.ACTION_NOTIFICATION_SETTINGS);
```
selvach, line 114: Remove this await. It causes issues.

src/core/push-notifications/PushNotificationService.ts
```diff
@@ -70,51 +70,48 @@ export default class PushNotificationService {
...
-        Analytics.track(events.NOTIFICATION_ENABLED);
-      } catch (error) {
-        // Fail silently
+      if (notify_backend && pushToken) {
+        // Send to our backend first and save locally only once successful.
+        await this.apiClient.updatePushToken(pushToken);
+        await this.savePushToken(pushToken);
       }
+    } catch (error) {
+      // Fail silently
```
selvach, line 104: Probably want to do something here...

selvach (changes requested):
Could you just merge in these unit tests on the `tokenNeedsRefreshing()` function, please!

https://github.com/zoe/covid-tracker-react-native/compare/selvach/force_new_token

src/core/push-notifications/PushNotificationService.ts
bionara: Fail Noisily @andyguest :)

bionara (dismissed):
Code looks good once Somesh's feedback is implemented

src/core/push-notifications/PushNotificationService.ts
andyguest: I'm maintaining the old error handling code here 😁 , but note that the silent failure still triggers a retry on next app opening.  Not sure if we can do much better than that?  Don't think we want to interrupt users with a pop-up for example, but maybe you'd prefer we did?

selvach (approved):

PR #1255, opened 2021-04-12
CircleCI Refactor

selvach:
## Why this PR? 

There are some things that bother me about the our CircleCI setup:

expo-cli: 
 - Warnings about using an outdated version
 - Warnings about the build version when doing an expo-cli publish
 - Installing it every time we do an OTA

builds:
 - We do two parallel builds every time, then decide whether or not to publish to store or do an OTA
 - Why do we need to build to do an OTA?
  
cocoapods:
 - Warning about mismatch between 1.10.0 and 1.10.1
 - pod dependencies do not seem to use the cache files. (Especially annoying when running tests). 

## This PR addresses the following 
- Only build after CI Approval for OTA vs Full Build. Prevents unnecessary Builds
- Install expo-cli into package.json (save 90seconds per OTA, as it takes expo-cli from the cache file rather installing it every time).
- Fix warnings with expo publish by adding build type explicitly
- Upgrades the macos executor image from `12.3.0` -> `12.4.0`. This upgrades coacoa pods from `1.10.0` -> `1.10.1`

## To Do
- Use the autobump versionID that Kenneth mentioned in.

## Merging this PR
- I'd only merge this PR when we want to upgrade our App Store Builds


.circleci/config.yml
```diff
@@ -122,7 +116,7 @@ commands:
       - run: curl -o- -L https://yarnpkg.com/install.sh | bash
       - run:
           name: Install Dependencies
-          command: yarn install --immutable
+          command: yarn install
```
andyguest, line 119: Can you explain why you removed this?  On the face of it, a zero-installs approach sounds superior to me.

.circleci/config.yml
```diff
@@ -171,8 +165,8 @@ commands:
   expo_ota_publish:
     steps:
       - set_expo_release_channel
-      - run: npx expo-cli@4.3.0 login -u $EXPO_USERNAME -p $EXPO_PASSWORD
-      - run: npx expo-cli@4.3.0 publish --non-interactive --max-workers 1 --release-channel ${RELEASE_CHANNEL}
+      - run: expo-cli login -u $EXPO_USERNAME -p $EXPO_PASSWORD
+      - run: expo-cli publish --non-interactive --max-workers 1 --release-channel ${RELEASE_CHANNEL} --target managed
```
andyguest, line 169: why is it target managed and not bare?

.circleci/config.yml
```diff
@@ -301,24 +295,6 @@ jobs:
       - prepare_deploy_env
       - expo_ota_publish
 
-  # ota-rollback:
```
andyguest, line 304: do we have an adequate rollback story already if we do get into bother?

package.json
```diff
@@ -70,6 +70,7 @@
     "expo": "39.0.0",
     "expo-analytics-amplitude": "^10.0.0",
     "expo-asset": "8.2.0",
+    "expo-cli": "4.3.5",
```
andyguest, line 73: (previously pegged to 4.3.0 so presumably the interim issue has been resolved?)

andyguest (commented):
I'm really not very clear on how the build process works so there's a dumb-ish question here.

> Only build after CI Approval for OTA vs Full Build. Prevents unnecessary Builds
Feels like the wrong way round, esp. if the build goes onto fail?  Why can't we make use of the output of the build step directly for the OTA?

package.json
selvach: Correct! We were getting warnings about being on an old version.

andyguest: I maybe put that pegging fix in because 4.3.1 didn't work, iirc 😂

.circleci/config.yml
andyguest: is it something like this?  https://github.com/expo/expo/issues/8293

.circleci/config.yml
selvach: I had wrongly assumed that the `--immutable` was the cause of the cache's being ignored. That's not the case. I've restored it.

.circleci/config.yml
selvach: Fixed. Thanks!

.circleci/config.yml
selvach: Yes: `expo publish:set --release-channel  --publish-id `

I thought there was a how to on notion, but can't find it. I'll write one up.

selvach: The previous CircleCI was completely untested. I think in the break the glass scenario, where we are not rolling forward, I'ld prefer to use the expo cli directly.

selvach:
You should only need to build (via fastlane) when pushing to the app store. 
An OTA only requires to bundle the TypeScript code and the assets.

andyguest (approved):
> You should only need to build (via fastlane) when pushing to the app store.
> An OTA only requires to bundle the TypeScript code and the assets.

Thanks - that makes a lot more sense to me now.  I still find the naming of the individual jobs within the workflows a bit confusing - think they could be more explicit about the what-and-why to aid future readers.

One quick example I find confusing being deploy-stage and deployed-stage.  pre-deploy-stage / post-deploy-stage perhaps?  But maybe not - I don't have the best mental model to suggest new names.  Happy enough to approve and we can revisit later if you agree.

PR #1266, opened 2021-04-20
Remove inconsistent shadows

selvach:
Removes inconsistent drop shadow on the DietScore card on the homepage. 

shadows


bionara (approved):
Nice detail!

PR #1267, opened 2021-04-20
Set push tokens to refresh daily

selvach:
Start refreshing the pushing notifications on the server faster. Charlie suggests checking every time the app starts is no bad thing. 

Woo: Nice to have unit tests!


andyguest (commented):
While I agree that Charlie said this was appropriate, it's still vulnerable and the fix should [also?] implement the listener here:

https://docs.expo.io/versions/latest/sdk/notifications/#addpushtokenlistenerlistener-pushtokenlistener-subscription

andyguest (approved):
Let's use this to clear out double notifications :-)

PR #1273, opened 2021-04-28
Add product catlalog description for backstage

selvach:
Metadata yaml file describing the service. 
This will be auto discovered by backstage.


bionara (commented):
Looks good

PR #1280, opened 2021-05-03
Hotfix Master: Removes Swedish Ad

selvach:
Now the seminar is finished, we can revert #1269

I am hot-fixing this change to the Swedish Thank You page straight to `master` - so we can OTA once the PR is approved. 
`develop` contains a bunch of changes the require a new App Store Build. 

### For Future PR:
I've raised a ticket to wire up the Swedish and US screens to the same headless CMS we use for UK. Then product can drive these changes without a code change. (This is the first time we've ever done something like this for Sweden).


aikewoody (approved):
The changes look exactly the opposite of #1269, so approved!

PR #1283, opened 2021-05-04
Switch to using deep links

selvach:
Switch to using deep links on the pop up that forces users to upgrade to the latest version. 

Inspired by this Stack Overflow:
https://stackoverflow.com/questions/62445835/how-to-open-directly-our-app-in-apple-store

### DIFF
--- a/ios/Covid/Info.plist
+++ b/ios/Covid/Info.plist
@@ -37,6 +37,10 @@
 	<string>4</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
+	<key>LSApplicationQueriesSchemes</key>
+    <array>
+        <string>itms-apps</string>
+    </array>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSAppTransportSecurity</key>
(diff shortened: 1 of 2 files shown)


ios/Covid/Info.plist
```diff
@@ -37,6 +37,10 @@
 	<string>4</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
+	<key>LSApplicationQueriesSchemes</key>
+    <array>
+        <string>itms-apps</string>
```
bionara, line 42: should this e items-apps?

bionara (approved):
approved with pending typo(?)

ios/Covid/Info.plist
selvach: `itms-apps` is correct. From what I have googled - it's the Apple Store specific url schema for deep linking to the App Store:

https://developer.apple.com/forums/thread/104915

https://stackoverflow.com/questions/47348083/what-is-the-difference-between-itms-app-and-https-app-store-link

bionara: ace

PR #1284, opened 2021-05-04
Removed the object destructuring in the ExternalCallout component.

aikewoody:
# Description

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [x] I have run `npm test`

### DIFF
--- a/src/components/ExternalCallout.tsx
+++ b/src/components/ExternalCallout.tsx
@@ -36,7 +36,6 @@ type ExternalCalloutProps = {
 };
 
 export const ExternalCallout: React.FC<ExternalCalloutProps> = (props) => {
-  const { calloutID, link, screenName, postClicked, canDismiss, orderIndex } = props;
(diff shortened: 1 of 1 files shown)


aikewoody:
I created this PR to have an open discussion about using object destructuring in this use-case. Let me start with writing down a list of pros and cons (feel free to add more!):

PROS
- In the line `{canDismiss && !props.isSharing && (` it's not clear where the variable `canDismiss` is coming from. Changing it to `{props.canDismiss && !props.isSharing && (` makes it easier readable.
- Saves an extra line of code by removing the variable assignments (can save a lot of lines of code if Prettier forces the assignment on separate lines). It is also more efficient but definitely not a noticeable difference for the user.
-

selvach (approved):
I agree this makes this more readable having `props.foo`. 

Overall better readability! WDYT @bionara ?

PR #1291, opened 2021-05-06
Add universal link permissions

selvach:
Following the instructions in here https://docs.expo.io/guides/linking/#universaldeep-links-without-a-custom-scheme to add Permissions for a universal link served from https://api.covidjoinzoe.com/launch/ to launch the app.


aikewoody (approved):
Based on our conversation on Slack it's approved!

PR #1304, opened 2021-05-12
Fix imports

selvach:


src/components/index.ts
```diff
@@ -3,7 +3,7 @@ export * from './animations';
 export * from './bars';
 export * from './buttons';
 export * from './cards';
-export * from './checkbox';
+export * from './Checkbox';
```
aikewoody, line 6: Good find! Scary that ESLint and TypeScript didn't find this...

aikewoody (approved):

PR #1321, opened 2021-05-21
Make a distinction in the mental health playback based on the users cohorts

aikewoody:
# Description

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/core/state/mental-health-playback/slice.ts
+++ b/src/core/state/mental-health-playback/slice.ts
@@ -3,132 +3,15 @@ import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 import { RootState } from '@covid/core/state/root';
 import { IMentalHealthPlayback, IInsightsDict } from '@covid/core/state/mental-health-playback/types';
 import { IInsight } from '@covid/types/mental-health-playback';
-
-// @todo
-const insights: IInsight[] = [
-  {
-    activity_name: 'sleeping_well_less',
-    answers: [
-      {
-        title: 'NO_CHANGE',
-        value: 0.46,
-        label: '46%',
-      },
-      {
-        title: 'LESS',
-        value: 0.37,
-        label: '37%',
-      },
-      {
-        title: 'MORE',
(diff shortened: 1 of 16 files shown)


src/core/state/mental-health-playback/slice.ts
```diff
@@ -33,6 +33,7 @@ const insights: IInsight[] = [
     direction: 'higher',
     level_of_association: 'strongly',
     segment: 'men_40',
+    user_answer: '',
```
bionara, line 36: Does this need a value?

src/features/AppCoordinator.ts
```diff
@@ -336,8 +336,8 @@ export class AppCoordinator extends Coordinator implements ISelectProfile, IEdit
     NavigatorService.navigate('MentalHealthChanges');
   }
 
-  goToMentalHealthStudyPlayback(user: any = {}) {
-    if (user.C) {
+  goToMentalHealthStudyPlayback(startupInfo: StartupInfo | undefined) {
+    if (startupInfo?.mh_insight_cohort === 'MHIP-v1-cohort_c') {
```
bionara, line 340: Worth setting the string as a const

bionara (commented):
Looks good so far (still a WIP).

src/core/state/mental-health-playback/slice.ts
aikewoody: When it's coming from the api yes, it served it's purpose as mockup data during development

src/features/AppCoordinator.ts
aikewoody: Good point

aikewoody: On the other hand because the object `startupInfo` is typed and `mh_insight_cohort` is defined as an enum typescript enforces using one of the three correct strings

PR #1330, opened 2021-05-25
Fix visual regression in branded buttons

selvach:
Fix the button text colour

Screenshot 2021-05-25 at 13 51 03


bionara (approved):
Very nice! I was looking at this just before i mentioned it, and was trying to work out a way of forcing the (white) style to the passed in <Text>... didn't think to just delete the node and pass a simple string lol!

PR #1349, opened 2021-06-03
Modal Events

selvach:
This PR removes the existing modal events sent to our backend on every view of every modal. It's replaced with two specific events that are called when the Mental Health Playback modal (the only currently active modal) is interacted with. (View Insights or Skip Insights).


aikewoody (approved):
Looks good to me

PR #1350, opened 2021-06-04
MHIP copy changes

selvach:


bionara (approved):
more is less
less is more
very profound

PR #1364, opened 2021-06-11
Removed hesitancy code that was still left behind.

aikewoody:
# Description

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/assets/lang/en.json
+++ b/assets/lang/en.json
@@ -1122,25 +1122,6 @@
       "name-missing": "Edit to add vaccine name",
       "not-logged": "Not yet logged",
       "edit-vaccine": "Edit vaccine"
-    },
-
-    "hesitancy": {
-      "title": "Would you accept a COVID-19 vaccine if offered?",
-      "check-all-that-apply": "Please tell us why (check all that apply):",
-      "vaccine-trial": "I took part on a vaccine trial",
-      "religious": "Religious reasons",
-      "philosophical": "Personal belief / philosophical reasons",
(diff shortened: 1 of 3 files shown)


src/features/anniversary/partials/ReportedEvent.tsx
```diff
@@ -14,8 +14,6 @@ function ReportedEvent({ reportedEvent }: IProps) {
         return 'loose-weight-1';
       case 'GENERAL':
         return 'location';
-      case 'HESITANCY':
```
selvach, line 17: We should keep this!

src/features/anniversary/types/index.ts
```diff
@@ -6,7 +6,6 @@ export type TBadge =
   | 'TEST_RESULTS'
   | 'VACCINATION_STATUS'
   | 'VACCINE_SIDE_EFFECTS'
-  | 'HESITANCY'
```
selvach, line 9: We should keep this!

src/features/anniversary/partials/ReportedEvent.tsx
aikewoody: That's what PR's are for! :)

src/features/anniversary/types/index.ts
aikewoody: That's what PR's are for! 👍

PR #1376, opened 2021-06-15
WIP: Use the correct Formik way of handling the submit.

aikewoody:
# Description

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist

- [ ] I have run `npm test`

### DIFF
--- a/src/components/buttons/Button.tsx
+++ b/src/components/buttons/Button.tsx
@@ -49,16 +49,13 @@ const styles = StyleSheet.create({
   darkLabel: {
     color: colors.brand,
   },
-
   lightLabel: {
     color: colors.white,
   },
-
   outlineContainer: {
     borderColor: colors.brand,
     borderWidth: 1,
   },
-
   outlineLabel: {
     color: colors.brand,
   },
--- a/src/components/buttons/SubmitButton.tsx
+++ b/src/components/buttons/SubmitButton.tsx
@@ -1,0 +1,17 @@
+import BrandedButton, { IProps as IBrandedButtonProps } from '@covid/components/buttons/branded-button';
+import { ErrorText } from '@covid/components/Text';
+import React from 'react';
+import { View } from 'react-native';
+
+interface IProps extends IBrandedButtonProps {
+  errorMessage?: string;
+}
+
+export function SubmitButton({ errorMessage, style, ...props }: IProps) {
+  return (
+    <View style={style}>
+      {errorMessage ? <ErrorText style={{ marginBottom: 8 }}>{errorMessage}</ErrorText> : null}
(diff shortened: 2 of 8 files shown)


bionara (dismissed):

selvach (commented):
Would you like to put a simple error message like this?
https://github.com/zoe/covid-tracker-react-native/blob/125ad1345d0d2f9b305d0adebc1a3770f25b2bd3/src/features/multi-profile/edit-profile/EditLocationScreen.tsx#L101

Or would you like to wait for a bigger refactor / clean up of error handling? 

O

selvach:
Tested this for double tapping the button -> two server calls. Looks good to me!

selvach:
Suggestions (can be done in another PR). 
 - Display an error message somewhere on the screen if you get an error from thrown by the API call. 
 - Display a spinner while the server submission is in progress. (i.e. before the screen transition). 

We could use this screen as an example of "how to do formik + server submission correctly.
The main thing is if we use Redux to capture the state, do we transition to the next screen before the server has responed with ("Okay" - I've saved your data)? 

CC: @juliatan @bionara

aikewoody:
> Suggestions (can be done in another PR).
> 
> * Display an error message somewhere on the screen if you get an error from thrown by the API call.
> * Display a spinner while the server submission is in progress. (i.e. before the screen transition).
> 
> We could use this screen as an example of "how to do formik + server submission correctly.
> The main thing is if we use Redux to capture the state, do we transition to the next screen before the server has responed with ("Okay" - I've saved your data)?
> 
> CC: @juliatan @bionara

I like that, this will serve as an example change for the other parts of the codebase 👌

juliatan:
> Suggestions (can be done in another PR).
> 
> * Display an error message somewhere on the screen if you get an error from thrown by the API call.
> * Display a spinner while the server submission is in progress. (i.e. before the screen transition).
> 
> We could use this screen as an example of "how to do formik + server submission correctly.
> The main thing is if we use Redux to capture the state, do we transition to the next screen before the server has responed with ("Okay" - I've saved your data)?
> 
> CC: @juliatan @bionara

Generally speaking (without having context around each form), my thinking process is:

- How important is it that we save the data from the user to the backend e.g. if used to calculate other results, or needed for future reporting, then we shouldn't allow user to proceed to next screen unless data is successfully saved.
- If it's not all that important, then probably makes sense to allow user to proceed so that they're not held up in any way. If this ultimately failed in the backend, it would be treated as lost data though (I don't think it's great UX to have them go back and fill in the form...but we could give them the option to and see if they take it up, though the user journey might be a little complex depending on the form).

PR #1385, opened 2021-06-16
Initial test of Font Awesome icons as a replacement for our SVGs

juliatan:
# Description

- Installed Font Awesome https://github.com/FortAwesome/react-native-fontawesome#installation
- Replaced edit profile and share app icons in the DrawerMenu as a test
- Ticket: https://www.notion.so/joinzoe/Replacing-SVG-Icons-with-FontAwesome-3b72e54fdc2342a88edf472392e40183

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 11 - 2021-06-16 at 18 12 38

Screenshot_1623864721

## Checklist

- [x] I have run `npm test`

- Upon the team agreeing that we want to proceed with Font Awesome icons library, I can go ahead and replace the rest of the SVGs and other icons.


juliatan:
@bionara What do you think about implementation? We can either do an explicit import of each icon into a particular component **every time** we want to use it. Or, we can implement a global library so we only import each icon once (which is what I've done here). Do you think it will get too messy if we import everything in `App.tsx`?

FYI @aikewoody @selvach

aikewoody:
I like the code changes it brings (deleting of svg components!) but to be honest the current icons look better. What other versions of these icons does font awesome have?

aikewoody:
In the past I have mostly used the library `vector-icons` https://github.com/oblador/react-native-vector-icons because it contains way more icon sets than just font awesome. @juliatan @bionara @selvach What do you think?

App.tsx
```diff
@@ -29,6 +31,8 @@ Sentry.init({
   environment: env.NAME,
 });
 
+library.add(faUser, faShareSquare);
```
aikewoody, line 34: I saw your question raised about this and it would be easier and cleaner to go for this:
`import { far } from '@fortawesome/free-regular-svg-icons';
...
library.add(far);
`
And now all the icons from the `far` icon set (font awesome regular) are loaded and can be used.

bionara: (font)Awesome work!

bionara:
Awesome work - closing as after discussions we decided against font awesome.

PR #1392, opened 2021-06-21
Refactor modal components to use react-native's Modal

juliatan:
# Description

- Refactor our modals to move away from Peter's route/navigator-based custom Modal, towards `react-native`'s Modal component.
- My proposal is that we keep the concept of a custom modal component (I've called it `ModalZoe` here, but would replace Peter's existing `Modal` once I'm done with all the refactoring) that utilises the `react-native` `Modal` and just sets the default styling so we have uniform modal across our code base.
- Rather than "navigating" to a modal (which is what we do now), the `ModalZoe` is either visible/hidden on the original page (in this case, the `Dashboard`). There's therefore no longer a need to add the modal components to the stack navigator, or have the concept of `goBack()`.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

N/A

## Checklist

- [ ] I have run `npm test`


bionara (approved):
Looks awesome! Was going to say we don't even need the ModalZoe wrapper, but that's cool as it lets us encapsulate the animation and so on. Nice one!

src/features/dashboard/DashboardScreen.tsx
```diff
@@ -29,6 +29,7 @@ import React, { useEffect, useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 import { useSelector } from 'react-redux';
 
+import MentalHealthPlaybackModal from '../mental-health-playback/MentalHealthPlaybackModal';
```
aikewoody, line 32: Would be good to always use absolute imports :)

src/features/dashboard/DashboardScreen.tsx
```diff
@@ -180,6 +182,15 @@ export function DashboardScreen({ navigation, route }: IProps) {
...
           </View>
         ) : null}
+
+        <ModalZoe
+          closeModalHandler={() => {
+            setMentalHealthPlaybackModalVisible(false);
+          }}
+          showModal={mentalHealthPlaybackModalVisible}
+        >
+          <MentalHealthPlaybackModal closeModalHandler={() => setMentalHealthPlaybackModalVisible(false)} />
```
aikewoody, line 192: How about moving the encapsulating `ModalZoe` component to the `MentalHealthPlaybackModal` component. This way the code follows a more DRY pattern (for example when another screen also uses the `MentalHealthPlaybackModal` you don't have to wrap it again with the `ModalZoe` component.

src/features/dashboard/DashboardScreen.tsx
juliatan: Ahh yep, gotcha. Thank you - I'll change this.

juliatan:
> Looks awesome! Was going to say we don't even need the ModalZoe wrapper, but that's cool as it lets us encapsulate the animation and so on. Nice one!

Yes, exactly, that was my thinking to try to keep all Modals standardised across our app. 🙂

src/features/dashboard/DashboardScreen.tsx
juliatan: Yes, upon reflection, I think your suggestion is a good one. Thank you - I'll make this change.

aikewoody (approved):

PR #1396, opened 2021-06-22
Delete AnniversaryModal component

juliatan:
# Description

Deleted AnniversaryModal as part of ongoing Modal refactor. 
Ticket: https://www.notion.so/joinzoe/Replacing-Modal-Components-9386b6fc74644ff2932b317a88286c1a

**Question: I've left the Anniversary Redux slice in place even though we're no longer dispatching actions to it.** I was a bit worried about Redux issues if I amended the state object without doing some kind of migration, so didn't want to risk it. I also wasn't sure how this works on mobile (can't seem to see a migration file in our code base, so perhaps it's done differently?). Can I speak to one of you guys about this pls? 😀

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

N/A

## Checklist (for submission)

- [x] ~Analytics/tracking events have been added (if needed)~ - Deleted analytics events that are no longer in use.

## Checklist (for reviewers)

Maybe the clean up of any Redux code which is no longer needed.


aikewoody:
@juliatan The whole `state.anniversary` is not used anymore in the codebase so it's safe to delete it

juliatan:
> @juliatan The whole `state.anniversary` is not used anymore in the codebase so it's safe to delete it

OK, thanks @aikewoody . As you mentioned, I'm getting this error in local environment, but it does also say "Unexpected keys will be ignored" in the last sentence, so hopefully that's ok!
Screenshot 2021-06-22 at 12 52 00

aikewoody:
@juliatan Yes exactly this was expected and will not do any harm on the production version of the app (we are certain of this because the slice `state.anniversary` in the redux state is not used anymore

aikewoody (approved):

src/features/anniversary/screens/index.ts
```diff
@@ -1,2 +1 @@
 export { default as Anniversary } from './Anniversary';
```
aikewoody, line 1: @juliatan Is the `Anniversary` screen still accessible for the user? If not, could it be deleted as well?

juliatan: It's still accessible from the ThankYou screen. 🙂

PR #1397, opened 2021-06-22
Upgrade VersionUpdateModal to react-native Modal

juliatan:
# Description

- Converted the `VersionUpdateModal` to utilise the `ModalZoe` component.
- Because this "should show modal" check happens in the `AppCoordinator`, immediately upon launch, and doesn't behave like a typical modal (i.e. with the ability to close it and go back to previous screen), I've left it in the navigation stack.
- I've also deleted Peter's modal component. As we can't have duplicate names in the same file (i.e. two `Modal`s), I've kept the name of our custom component `ModalZoe`.
- Ticket: https://www.notion.so/joinzoe/Replacing-Modal-Components-9386b6fc74644ff2932b317a88286c1a

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

N/A

## Checklist (for submission)

- ~[ ] Analytics/tracking events have been added (if needed)~

## Checklist (for reviewers)

N/A


aikewoody:
To be honest I think we should try to get this modal also out of the navigation stack. Otherwise we end up with two different ways and that will get confusing

juliatan:
> To be honest I think we should try to get this modal also out of the navigation stack. Otherwise we end up with two different ways and that will get confusing

I thought about moving this to the DashboardScreen... but am worried about multiple modals popping up at the same time. What do you think?

bionara:
> > To be honest I think we should try to get this modal also out of the navigation stack. Otherwise we end up with two different ways and that will get confusing
> 
> I thought about moving this to the DashboardScreen... but am worried about multiple modals popping up at the same time. What do you think?

yep agree. Happy to merge it as a progressive enhancement though, and clear up the PRs!

aikewoody:
> > To be honest I think we should try to get this modal also out of the navigation stack. Otherwise we end up with two different ways and that will get confusing
> 
> I thought about moving this to the DashboardScreen... but am worried about multiple modals popping up at the same time. What do you think?

It's better to make sure such a thing can't happen, but that could be for another PR :)

juliatan:
> > > To be honest I think we should try to get this modal also out of the navigation stack. Otherwise we end up with two different ways and that will get confusing
> > 
> > 
> > I thought about moving this to the DashboardScreen... but am worried about multiple modals popping up at the same time. What do you think?
> 
> It's better to make sure such a thing can't happen, but that could be for another PR :)

OK, thanks both. Let's do it progressively then. If someone can approve, I can merge and continue looking into it. 😄

selvach (approved):

aikewoody (approved):

PR #1398, opened 2021-06-22
Update style guide for how we define components

juliatan:
I had been wondering this myself, so thought I'd document it. This is based on Somesh's recent PR: #1394 .


selvach (dismissed):

selvach:
Thank you for doing this :)

CONTRIBUTING.md
```diff
@@ -36,6 +36,17 @@ Now that we've excluded a few things, let's discuss what would be great contribu
 We're trying to standardise how we write our code, to make it as easy as possible for others to contribute. Here are some of the standards we use:
 
 - Airbnb's eslint plugin with Typescript: [`eslint-config-airbnb-typescript`](https://github.com/iamturns/eslint-config-airbnb-typescript)
+- Define components using the `function` syntax (rather than `const` arrow functions), and destructure props e.g.
+
+```jsx
+export default function CovidTestConfirmScreen({ navigation, route }: IProps) {
```
aikewoody, line 42: Not sure about the destructuring of the props. It means that every time a prop is added/removed/changed from a component you also have to change the function parameters. Whereas if you keep is like `(props: IProps)` you won't need to do that

aikewoody:
@juliatan It's great that you document so many things! (something I have to work on myself)

CONTRIBUTING.md
bionara: I'm with @aikewoody on this one, then access them as props.whatevs in the component - makes it easier to see that it's come from props and not a component (or external!) const also.

bionara (dismissed):
Great work! Mentioned the prop think Aike raised but happy to merge

CONTRIBUTING.md
juliatan: Lols, so what's the conclusion? I prefer being more explicit about props and destructuring (rather than seeing `props` everywhere). Happy to go with whatever though.

aikewoody: Also a good point that when you keep doing `foo={props.items}` you know the value is coming from the properties. Whereas `foo={items}` doesn't tell you anything, could be from an import or const declaration in the render or from somewhere else.

selvach (approved):

PR #1410, opened 2021-06-24
Form validation refactor

juliatan:
# Description

Add form validation to AboutYourHealthScreen form.
No specific ticket, logging finished refactors here: https://www.notion.so/joinzoe/Fun-with-Forms-d66ad33a88694d8aaa86777fbf4a75b8#22636d7c049b4132a051c3a5eea4b8fc

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 11 - 2021-06-24 at 17 03 47
Simulator Screen Shot - iPhone 11 - 2021-06-24 at 17 03 57
Simulator Screen Shot - iPhone 11 - 2021-06-24 at 17 04 02

## Checklist (for submission)

- ~[ ] Analytics/tracking events have been added (if needed)~

## Checklist (for reviewers)

Part of bigger Form refactor.

### DIFF
--- a/src/features/patient/YourHealthScreen.tsx
+++ b/src/features/patient/YourHealthScreen.tsx
@@ -1,4 +1,5 @@
 import { BrandedButton } from '@covid/components';
+import { FormWrapper } from '@covid/components/Forms';
 import { GenericTextField } from '@covid/components/GenericTextField';
 import { RadioInput } from '@covid/components/inputs/RadioInput';
 import ProgressStatus from '@covid/components/ProgressStatus';
@@ -20,7 +21,6 @@ import { stripAndRound } from '@covid/utils/number';
 import { RouteProp } from '@react-navigation/native';
 import { StackNavigationProp } from '@react-navigation/stack';
 import { Formik, FormikProps } from 'formik';
-import { Form } from 'native-base';
 import * as React from 'react';
 import { View } from 'react-native';
 import * as Yup from 'yup';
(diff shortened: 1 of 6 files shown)


src/features/patient/YourHealthScreen.tsx
```diff
@@ -384,8 +396,10 @@ export default class YourHealthScreen extends React.Component<HealthProps, State
                     <ValidationError error={i18n.t('validation-error-text')} />
                   ) : null}
                 </View>
-                <BrandedButton onPress={props.handleSubmit}>{i18n.t('next-question')}</BrandedButton>
-              </Form>
+                <BrandedButton enable={props.isValid && props.dirty} onPress={props.handleSubmit}>
```
aikewoody, line 399: Is it necessary to also include `&& props.dirty` into the check?

juliatan: Yes it is, otherwise the button will be enabled when a user first lands on the screen. It's only when the user clicks into a form input that validation runs, and `isValid` is checked.

aikewoody: Got it! So, if in the future all the inputs in this form change to optional the check on `props.dirty` should be deleted because otherwise the button wouldn't be enabled on first load without the user interacting with the form

aikewoody (approved):

src/features/patient/YourHealthScreen.tsx
juliatan: Yup that's correct!

PR #1412, opened 2021-06-28
Fixed About you form back error

bionara:
# Description
Fixes the error state for "About you" form on back.
Note that this form, and others, should be refactored to be functional components, in which case we can better handle any back/navigation events with useEffect.

https://www.notion.so/joinzoe/017c3de591604a7b89d0b69e3c37b279?v=0133357df3ed42e1bd42c1e0d7afcf45&p=b928c836e318483aa47ff045eb9417a3

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

### DIFF
--- a/src/features/patient/AboutYouScreen.tsx
+++ b/src/features/patient/AboutYouScreen.tsx
@@ -343,7 +343,6 @@ export default class AboutYouScreen extends React.Component<AboutYouProps, State
 
         <Formik
           validateOnChange
-          validateOnMount
           initialValues={this.props.route.params?.editing ? this.getPatientFormValues() : getInitialFormValues()}
           onSubmit={(values: IAboutYouData) => {
             return this.handleUpdateHealth(values);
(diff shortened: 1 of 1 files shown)


src/features/patient/AboutYouScreen.tsx
```diff
@@ -458,7 +457,7 @@ export default class AboutYouScreen extends React.Component<AboutYouProps, State
                   ) : null}
                 </View>
 
-                <BrandedButton enable={props.isValid} loading={props.isSubmitting} onPress={props.handleSubmit}>
+                <BrandedButton enable={props.isValid && props.dirty} onPress={props.handleSubmit}>
```
aikewoody, line 460: Don't we want to show the loading spinner?

bionara: Sorry should've added to the PR comment - as mentioned around the useEffect, we don't actually have a way of setting isSubmitting on navigate back, so it'll always be locked in that state once set. This seems to be the best approach until we get a global solution.

juliatan (approved):

aikewoody (approved):

PR #1421, opened 2021-06-29
Add required prop to fields in PreviousExposure form

juliatan:
# Description

Add required props to the PreviousExposure to COVID form.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 11 - 2021-06-29 at 16 22 41

## Checklist (for submission)

- [ ] Analytics/tracking events have been added (if needed)

## Checklist (for reviewers)

Need to refactor to a functional component so that we can reset the form if a user clicks back on navbar after submitting. 
Ticket: https://www.notion.so/joinzoe/Product-Eng-Board-ca9e8b1b8926481f8c94cb01a88d567d?p=5c5011dd2c9f4b86a01f1db55820f8bd

### DIFF
--- a/src/features/patient/PreviousExposureScreen.tsx
+++ b/src/features/patient/PreviousExposureScreen.tsx
@@ -1,5 +1,6 @@
 import { BrandedButton } from '@covid/components';
 import { CheckboxItem, CheckboxList } from '@covid/components/Checkbox';
+import { FormWrapper } from '@covid/components/Forms';
 import { GenericTextField } from '@covid/components/GenericTextField';
 import { RadioInput } from '@covid/components/inputs/RadioInput';
 import ProgressStatus from '@covid/components/ProgressStatus';
@@ -18,7 +19,7 @@ import { stripAndRound } from '@covid/utils/number';
 import { RouteProp } from '@react-navigation/native';
 import { StackNavigationProp } from '@react-navigation/stack';
 import { Formik } from 'formik';
-import { Form, Item, Label } from 'native-base';
+import { Item, Label } from 'native-base';
 import * as React from 'react';
 import { StyleSheet, View } from 'react-native';
 import * as Yup from 'yup';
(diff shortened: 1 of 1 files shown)


src/features/patient/PreviousExposureScreen.tsx
```diff
@@ -33,7 +34,7 @@ interface IYourHealthData {
 
 const initialFormValues = {
   classicSymptoms: 'no',
-  pastSymptomsChanged: 'no',
+  pastSymptomsChanged: 'much_better',
```
aikewoody, line 37: Is this change intentional?

src/features/patient/PreviousExposureScreen.tsx
```diff
@@ -290,8 +296,10 @@ export default class PreviousExposureScreen extends React.Component<HealthProps,
                   ) : null}
                 </View>
 
-                <BrandedButton onPress={props.handleSubmit}>{i18n.t('next-question')}</BrandedButton>
-              </Form>
+                <BrandedButton enable={props.isValid} onPress={props.handleSubmit}>
```
aikewoody, line 299: Should we also check for `props.dirty` like on the other forms?

src/features/patient/PreviousExposureScreen.tsx
juliatan: Yep, this is intentional - `no` doesn't exist as an option.

src/features/patient/PreviousExposureScreen.tsx
juliatan: It's not needed in this case because there's just one text input field here which is optional. :)

aikewoody: How about this line?
```
                      <GenericTextField
                        required
                        formikProps={props}
                        keyboardType="numeric"
                        label={i18n.t('label-past-symptoms-days-ago')}
                        name="pastSymptomsDaysAgo"
                      />
```

juliatan: It's hidden unless a specific yes/no field is toggled which means that form will have to be dirty by default to be valid.
Upon form load, there's only one yes/no toggle which auto defaults to "no", so theoretically, the form should be enabled, even if user doesn't "dirty" it.

aikewoody (approved):

PR #1447, opened 2021-07-05
Restructuring "reconsent" slice and save disease preferences to backend

juliatan:
# Description

https://www.notion.so/joinzoe/Re-consent-Flow-Use-Formik-state-to-manage-disease-preferences-d2e26c534ea143a385f044a09e4aeefc

https://www.notion.so/joinzoe/Product-Eng-Board-ca9e8b1b8926481f8c94cb01a88d567d?p=5cd3749b554c4dfbaad17d5e1667a59f

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

n/a

## Checklist (for submission)

- [ ] Analytics/tracking events have been added (if needed)

## Checklist (for reviewers)

Wait for backend migrations to check that disease preferences are being saved.

### DIFF
--- a/src/core/state/reconsent/slice/index.ts
+++ b/src/core/state/reconsent/slice/index.ts
@@ -12,6 +12,7 @@ const reconsentSlice = createSlice({
   reducers: {
     updateDiseasePreferences: (state, action: PayloadAction<TDiseasePreferencesData>) => {
       return {
+        ...state,
         diseasePreferences: action.payload,
       };
     },
--- a/src/core/state/reconsent/slice/index.ts
+++ b/src/core/state/reconsent/slice/index.ts
@@ -1,30 +1,18 @@
-import { TDiseasePreferencesData } from '@covid/core/state/reconsent/types';
+import { IReconsent, TDiseasePreferencesData } from '@covid/core/state/reconsent/types';
 import { RootState } from '@covid/core/state/root';
-import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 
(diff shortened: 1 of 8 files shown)


src/core/state/reconsent/slice/index.ts
```diff
@@ -1,30 +1,18 @@
...
 
 const reconsentSlice = createSlice({
-  initialState: initialStateDiseasePreferences,
+  initialState: initialStateReconsent,
   name: 'ReconsentState',
   reducers: {
     updateDiseasePreferences: (state, action: PayloadAction<TDiseasePreferencesData>) => {
       return {
-        ...action.payload,
+        diseasePreferences: action.payload,
```
aikewoody, line 15: You should create a new state object from the current state and then override the `diseasePreferences`, like this for example:
```
return {
  ...state,
  diseasePreferences: action.payload,
}
```

aikewoody (changes requested):
Looks good! Only the `updateDiseasePreferences` reducer needs a fix

src/core/state/reconsent/slice/index.ts
juliatan: Thanks Aike. I actually don't want the existing state at all, because the payload only includes the current list of disease preferences chosen.

If existing state includes e.g.` mental_health: true` but user changes their mind later and toggles it off and chooses` womens_health: true` instead, if we keep existing state with `...state`, we'd have `mental_health: true` and `womens_health: true`.

juliatan: O shoot, yea. Sorry forgot to change this since I amended the state object. Thanks!

aikewoody (approved):

PR #1487, opened 2021-07-14
WIP: Save the disease preferences only in the global redux state.

aikewoody:
# Description

@juliatan Sorry to come with this almost at the end of the implementation of the feature but I feel there are a few things that could be improved:

- The `DiseaseCard` component doesn't need a local state and overriding the `onPress` property.
- The global redux state and local component state shouldn't be mixed together like in `ReconsentDiseasePreferencesScreen`.
- Use a selector in `ReconsentDiseaseSummaryScreen` to get the chosen diseases.

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)


juliatan:
Thanks for improvements @aikewoody ! 🙂  Yeh agree it's cleaner to just use global state. It goes against what Christina agreed regarding immediately persisting disease choice, but I think it's fine considering it's unlikely users will close app on the disease preferences screen.

I noticed a bug where neurological conditions card seems missing at times when navigating between first and second screens - see gif.
Jul-14-2021 14-46-43 2

aikewoody:
@juliatan Yes exactly because it goes a bit against what Christina said I was hesitating to do this but on a technical level it a big improvement so hopefully nobody will notice haha

Good spot about that missing card! Very strange, I'll look into it

juliatan (approved):

PR #1501, opened 2021-07-19
Ben/active case linebreak

bionara:
# Description

Quick fix for line breaking number on homepage: added `numberOfLines` to the Text wrapper, and re-jigged the layout slightly to be more uniform. Screenshot has ",123" added to the number manually to demo the bug doesn't occur. We could make the font smaller if we exceed these numbers, but if that happens, we probably have bigger things to worry about than a line break un the UI...!!

https://www.notion.so/joinzoe/Fix-Wrapping-of-Active-Cases-e5cd0ef52b3e43b7a5bf731ec162e2a6

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

image

## Checklist (for submission)

## Checklist (for reviewers)


src/components/cards/estimated-case/EstimatedCaseCard.tsx
```diff
@@ -44,15 +44,15 @@ export function EstimatedCaseCard({
         </View>
       </View>
       <View style={styles.metricRow}>
-        <View style={styles.metricBox}>
-          <Text textClass="h1Regular">{leftMetric}</Text>
+        <View >
```
aikewoody, line 47: Needs a `yarn prettier` :)

src/components/typography/text/index.tsx
```diff
@@ -32,6 +32,7 @@ export interface IProps {
   textAlign?: TTextAlign | undefined;
   textDecorationLine?: TTtextDecorationLine | undefined;
   textClass?: TTextClass;
+  numberOfLines?: number | undefined;
```
aikewoody, line 35: It's probably a copy paste from above but this should def be `numberOfLines?: number;`

bionara: updated

src/components/cards/estimated-case/EstimatedCaseCard.tsx
bionara: prettiness ++

aikewoody (approved):

selvach:
> We could make the font smaller if we exceed these numbers, but if that happens, we probably have bigger things to worry about than a line break un the UI...!!

😂

PR #1503, opened 2021-07-19
COVID Test UI update - Add test screen

juliatan:
# Description

Ticket (part 1): https://www.notion.so/joinzoe/COVID-Test-Screen-UI-update-Add-New-Test-Screen-a4fe50cf67eb44b79af00da3ace46652

There are already quite a lot of changes, so eventhough the ticket is not completed, I think this is a good place to request for a review. Here's the Figma: https://www.figma.com/file/7TEcX7f6rw3Z0WYR4TpX3R/COVID-Symptom-Study-(App)---MASTER?node-id=12017%3A157720

**Changes:**
-  Convert `src/features/covid-tests/CovidTestDetailScreen.tsx` to a functional component
-  Amend title copy to "Add new test for COVID-19"
-  Remove progress bar underneath title
-  Add switch for approximate date range
-  Leave calendar view open even when a user has selected a date
-  Change copy of button to "Add test"

**Still to be completed:**
- Conditional questions based on type of test chosen
- Translations
- Updating the various types of tests available - blocked. See Notion for more details.
- There are some design variations for the Switch and Calendar. I've added this as a TBD in Notion for now. It's not a simple fix, so propose that we leave design tweaks to the end once the functionality is completed.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 11 - 2021-07-19 at 20 39 38
Simulator Screen Shot - iPhone 11 - 2021-07-19 at 20 39 44

## Checklist (for submission)

## Checklist (for reviewers)

Yes - see above.
Also suggest we look at implementing something like a [React Query](https://react-query.tanstack.com/) to reduce repetition around our API calls as I find myself repeating the same things like setting loading, setting error states etc. Will add to refactor list.


src/features/covid-tests/CovidTestDetailScreen.tsx
```diff
@@ -161,93 +144,98 @@ export default class CovidTestDetailScreen extends React.Component<CovidProps, S
...
+                  test={test}
+                />
+                <CovidTestResultQuestion
+                  formikProps={props as unknown as FormikProps<ICovidTestResultData>}
+                  test={test}
+                />
+                {/* <CovidTestIsRapidQuestion
+                  formikProps={props as unknown as FormikProps<ICovidTestIsRapidData>}
+                  test={test}
+                /> */}
```
aikewoody, line 211: Is this commented out because it's WIP?

src/components/Switch.tsx
```diff
@@ -0,0 +1,46 @@
+import { Text } from '@covid/components';
```
aikewoody, line 1: Because this switch is used as an input in a form let's move it to `src/components/inputs/`

src/features/covid-tests/CovidTestDetailScreen.tsx
```diff
@@ -161,93 +144,98 @@ export default class CovidTestDetailScreen extends React.Component<CovidProps, S
...
+    >
+      <Header>
+        <HeaderText>
+          {i18n.t(testId ? 'covid-test.page-title-detail-update' : 'covid-test.page-title-detail-add')}
+        </HeaderText>
+      </Header>
+
+      <Formik
+        validateOnMount
+        initialValues={{
```
aikewoody, line 178: (just a suggestion) While you're at it can this be improved by memoizing it? Not sure though how the individual `initialFormValues(test)` work and if it's suitable for memoization but if so that would be great!

aikewoody:
Looks great! Good job on converting that big class component to a FC!

aikewoody (commented):
Looks great, good job on converting that big class component to a FC! Just a few comments to be resolved and then it's GTG!

src/features/covid-tests/CovidTestDetailScreen.tsx
bionara: Good thought, and worth thinking similar for any other refactors that could benefit from this!

src/features/covid-tests/CovidTestDetailScreen.tsx
juliatan: Yup exactly :)

src/features/covid-tests/CovidTestDetailScreen.tsx
juliatan: Thanks Aike, that's a good thought. I've looked into it quickly, and I think if we really did want to memoise our forms, we'd do this at an individual field level.

i.e. I think we'd wrap e.g. our TextInput field, checkbox input etc. with React.memo. This means that even if Formik parent rerenders because its state is changing, child components don't rerender unless necessary.

I'm a little wary of implementing this across the board as it might have unintentional consequences, especially because a bunch of our forms contain fields that are conditional on other fields. We may have to look at it on a form by form basis and consider overall whether it's a good idea. Definitely makes sense if a form has say 30 fields and there's perceivable lag... otherwise, we'd need to weigh pros and cons. Curious to hear what you guys think.

src/components/Switch.tsx
juliatan: Arguably, it's just a dumb UI component, in that the true/false value from this switch doesn't get saved as a form answer. This component can therefore be used outside a form, and its sole purpose is toggling condition to show/hide other components.

PR #1521, opened 2021-07-21
Tests list view with tab navigator

juliatan:
# Description

Notion: https://www.notion.so/joinzoe/COVID-Test-Screen-UI-update-Add-New-Test-Screen-a4fe50cf67eb44b79af00da3ace46652

Note: Other tab only shown if "other" tests exists. Otherwise, just 3 tabs are shown.
Have checked the design of the 4 visible tabs view against numerous device sizes and with Sol, and she's happy with it.
The list itself is scrollable, depending on the size of the screen.

To do:
- Translations are still TBD.

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Screenshot 2021-07-21 at 16 28 28
Screenshot 2021-07-21 at 16 49 02
Screenshot 2021-07-21 at 16 54 14
Screenshot 2021-07-21 at 17 03 14
Screenshot 2021-07-21 at 17 05 26
Simulator Screen Shot - iPhone 11 - 2021-07-21 at 17 13 57

## Checklist (for submission)

## Checklist (for reviewers)


src/features/covid-tests/CovidTestTabbedListsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const otherTests = props.covidTests.filter((item: CovidTest) => isOtherTest(item.mechanism));
+
+  // 4.5 to achieve the cut off effect of "other" tab
+  const tabWidth = Math.max(
+    85,
+    otherTests.length ? windowDimensions.width / 4.5 : (windowDimensions.width - GUTTER) / 3,
+  );
+
+  return (
+    <Tab.Navigator
```
aikewoody, line 86: Would be great to have this in a separate component to be reused easily in the future!

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Tab.Screen children={() => <CovidListByType covidTests={antibodyTests} />} name="Antibody" />
+
+      {otherTests.length ? (
+        <Tab.Screen children={() => <CovidListByType covidTests={otherTests} />} name="Other" />
+      ) : null}
+    </Tab.Navigator>
+  );
+}
+
+const styles = StyleSheet.create({});
```
aikewoody, line 119: Not needed anymore :)

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export default function CovidTestTabbedListsScreen(props: IProps) {
+  const windowDimensions = useWindowDimensions();
+
+  const lateralFlowTests = props.covidTests.filter((item: CovidTest) =>
+    isLateralFlowTest(item.mechanism, item.is_rapid_test),
+  );
+  const pcrTests = props.covidTests.filter((item: CovidTest) => isPcrTest(item.mechanism, item.is_rapid_test));
+  const antibodyTests = props.covidTests.filter((item: CovidTest) => isAntibodyTest(item.mechanism));
+  const otherTests = props.covidTests.filter((item: CovidTest) => isOtherTest(item.mechanism));
```
aikewoody, line 77: Would be better for performance if these list operations are memoized by wrapping them in a `React.useMemo`

juliatan: Yes true... I need to remember to keep asking myself this. Never quite sure on appropriate use of `useMemo`.

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
juliatan: Ah thanks! Good spot.

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
juliatan: Thanks Aike! Yea I thought about this but TBH, I am not hugely in favour of abstracting components, until there's at least a 2nd use case. Reasons:
1. The 2nd use case might have specific needs of its own which means 2nd developer having to pick through abstracted component to "make it fit".
2. There might never be a 2nd use case.

If you guys feel strongly about this, then let me know and I'm happy to be dissuaded.

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+        style: { marginBottom: 12 },
+        tabStyle: { paddingHorizontal: 6, paddingVertical: 14, width: tabWidth },
+      }}
+    >
+      <Tab.Screen children={() => <CovidListByType covidTests={lateralFlowTests} />} name="Lateral" />
+      <Tab.Screen children={() => <CovidListByType covidTests={pcrTests} />} name="PCR" />
+      <Tab.Screen children={() => <CovidListByType covidTests={antibodyTests} />} name="Antibody" />
+
+      {otherTests.length ? (
+        <Tab.Screen children={() => <CovidListByType covidTests={otherTests} />} name="Other" />
```
aikewoody, line 113: The react-navigation documentation says to use `component` and not `children`, is there a specific reason you went for this? I am asking because now the ESLint rule `react/no-children-prop` has to be disabled which is something we should prevent from doing

juliatan: Yup, it's because you can't otherwise pass a prop into your component. i.e. can't pass `covidTests` in.

aikewoody:
The tab looks great! There is a warning that's logged multiple times in runtime that need to be solved:
`VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead.`

juliatan:
> The tab looks great! There is a warning that's logged multiple times in runtime that need to be solved:
> `VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead.`

What on earth...? 😕 Thanks for pointing out - I'll look into this.

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
aikewoody: The property you are passing through `() => <CovidListByType covidTests={antibodyTests} />` is also a component because it's a function that returns a JSX element. So you could change `children` to `component`

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
aikewoody: Good question, a good rule of thumb is if there is a list operator involved always memoize it

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
aikewoody: I do get your points, what I see happening in practice especially in our codebase (mostly from the previous developers) is that in the end when a 2nd use case comes along it's not extracted into a generic component but reimplemented again.
For example, in the future when a different team is working on the project they would consider this screen legacy and be very hesitant to change it and big chance they don't extract this specific code into a component that can be reused.

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
juliatan: Mmm, I don't think it worked when I tried it but let me double check again. If I can make it work with `component`, I'll change it.

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
aikewoody: Having said that, if it's a lot of work for you to extract this then it's not worth the time but otherwise it's good practice to make generic components (also to enforce consistency)

juliatan: Yeh, gotta say it also makes more sense to abstract out if we have something like Storybook to show new devs what existing components there are as well.

I do take your point though.

juliatan: hmm, let me add a ticket and once I'm done with the rest of the feature, can come back to it!

src/features/covid-tests/CovidTestTabbedListsScreen.tsx
juliatan: Yea, so if we use `component` instead of `children`, we get a perf warning.
Screenshot 2021-07-22 at 13 12 20

bionara (approved):
Looks awesome!

PR #1528, opened 2021-07-26
Finalised the second a/b test for the mental health playback.

aikewoody:
# Description

https://www.notion.so/joinzoe/Mental-health-playback-A-B-test-7c6ea57a7d4e4bb48dbe211ea96297bf

https://www.figma.com/file/7TEcX7f6rw3Z0WYR4TpX3R/COVID-Symptom-Study-(App)---MASTER?node-id=12512%3A132033

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

- [x] Analytics/tracking events have been added (if relevant)

## Checklist (for reviewers)


selvach:
I'm okay with all the refactors in this PR!

We need to discuss the implementation of `trackMHP()` and `trackScreenViewMHP`

This article (https://help.amplitude.com/hc/en-us/articles/207108327 ) describes the difference between:
 - event_properties (set on a call to `Amplitude.track()`). Needs to be passed on every event if you want to segement and create charts. 
 - user_properties (set on a call to `Amplitdue.identify()`). After setting a user_property ALL events for that user from that point onwards can be filtered by this property when creating charts.

aikewoody:
@selvach Sounds like using `identify` is the easier and also preferred option in this case. You agree on changing `trackMHP` and `trackScreenViewMHP` to the standard functions we already have?

selvach:
Yes! I think that's the preferred approach @aikewoody. 

If you look at the `Experiments.ts` code - at the point you get the user's cohort / group for the experiment you make the `identify()` call. 

https://github.com/zoe/covid-tracker-react-native/blob/bd7f9fcc67b2b220d7f9018e4992253f6433fb06/src/core/Experiments.ts#L34

aikewoody:
> Yes! I think that's the preferred approach @aikewoody.
> 
> If you look at the `Experiments.ts` code - at the point you get the user's cohort / group for the experiment you make the `identify()` call.
> 
> https://github.com/zoe/covid-tracker-react-native/blob/bd7f9fcc67b2b220d7f9018e4992253f6433fb06/src/core/Experiments.ts#L34

Thanks for this suggestion, it made the code a lot simpler!

selvach (approved):
I gave this a good test! Looks great :)

PR #1530, opened 2021-07-26
Save antibody type and test_performed_by answers to match backend

juliatan:
# Description

https://www.notion.so/joinzoe/017c3de591604a7b89d0b69e3c37b279?v=0133357df3ed42e1bd42c1e0d7afcf45&p=a4fe50cf67eb44b79af00da3ace46652

This PR covers:
- Checks to ensure we're saving the antibody type to the backend.
- Made the test_performed_by question a separate Formik field from trained_worker, since we decided to save it as a separate field in the backend.
- Upon adding/update a particular type of test e.g. PCR, when user is taken back to the List view, the tab shown auto defaults to the type of test they just added/updated. 

Note: there are now 2 similar questions.
- Old: "Did a trained worked swab you?"
- New: "Who performed the test?"
- New question is only shown for PCR tests. Old question is only shown if someone amends an old test THAT IS NOT a PCR test.

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Jul-26-2021 20-58-16

## Checklist (for submission)

## Checklist (for reviewers)


src/features/ScreenParamList.tsx
```diff
@@ -53,7 +54,12 @@ export type TScreenParamList = {
 
   // Assessment screens
   HealthWorkerExposure: { assessmentData: TAssessmentData };
-  CovidTestList: { assessmentData: TAssessmentData; tests?: TCovidTest[] };
+  CovidTestList: {
+    assessmentData: TAssessmentData;
+    tests?: TCovidTest[];
+    mechanism: ECovidTestMechanismOptions;
+    is_rapid_test: boolean;
```
aikewoody, line 61: Are these navigation params really required?

src/features/covid-tests/CovidTestListScreen.tsx
```diff
@@ -15,6 +15,8 @@ import { colors } from '@theme';
 import * as React from 'react';
 import { StyleSheet, useWindowDimensions, View } from 'react-native';
 
+import { getTestType } from './helpers';
```
aikewoody, line 18: Absolute imports :)

src/features/covid-tests/CovidTestListScreen.tsx
```diff
@@ -71,6 +73,10 @@ export default function CovidTestListScreen(props: IProps) {
 
   const currentPatient = assessmentCoordinator.assessmentData?.patientData?.patientState;
 
+  const showTab = props.route.params?.mechanism
+    ? getTestType(props.route.params.mechanism, props.route.params.is_rapid_test)
+    : null;
```
aikewoody, line 78: It seems like `Lateral` is the default, better to change `: null;` with `: 'Lateral';` and then `showTab={showTab}`

src/features/covid-tests/fields/CovidTestMechanismQuestion.tsx
```diff
@@ -109,7 +118,10 @@ export const CovidTestMechanismQuestion: ICovidTestMechanismQuestion<IProps, ICo
         />
       )}
 
-      {formikProps.values.mechanism === 'pcr' && (
+      {test !== undefined &&
```
aikewoody, line 121: The `test !== undefined` bit is redundant because this check `test?.trained_worker !== null` is already filtering that out :)

src/features/covid-tests/fields/CovidTestMechanismQuestion.tsx
```diff
@@ -109,7 +118,10 @@ export const CovidTestMechanismQuestion: ICovidTestMechanismQuestion<IProps, ICo
...
+      test?.trained_worker !== null &&
```
aikewoody, line 122: What if `test?.trained_worker` is `undefined`? then this statement is true, maybe better to check for a truthy value? e.g. `!!test?.trained_worker &&`

src/features/ScreenParamList.tsx
juliatan: Yep - I need it to basically check what type of test was just updated / added so I can open that tab view directly upon user being redirected back to list view.

src/features/covid-tests/CovidTestListScreen.tsx
juliatan: Yes, true.

src/features/covid-tests/fields/CovidTestMechanismQuestion.tsx
juliatan: I think it's needed for the Typescript type guarding in line 124... Typescript otherwise complains about value potentially being undefined and I wanted to keep `TCovidTest` strict and not add `undefined` unnecessarily.

src/features/covid-tests/fields/CovidTestMechanismQuestion.tsx
juliatan: I think I checked if this was possible from backend and doesn't seem to be... but I'll add since no harm! 😄

PR #1535, opened 2021-07-27
Fix for trendlines.

selvach:
# Description

Notion Ticket: 
https://www.notion.so/joinzoe/Detailed-Trendline-not-displaying-e73bb2f279e1443db40ac3db6813aa53

The detailed trendline uses a different charting library to plot the graphs than on the HomeScreen.

For some reason, this library does not like values of 0 when the min of y-axis is also set to zero. Put it another way - it doesn't like it if the trendline touches the x-axis.

I don't know if the change in methodology made it more common for the 0 values, or perhaps people are looking at this graph more.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Untitled

## Checklist (for submission)

## Checklist (for reviewers)


juliatan (approved):

src/components/stats/TrendLineChart.tsx
```diff
@@ -133,7 +133,7 @@ export function TrendLineChart({ filter, viewMode }: IProps) {
           payload: {
             data: timeseriesSorted,
             max: Math.max(...values),
-            min: Math.min(...values),
+            min: Math.min(...values) -1, // The "Explore" chart does not like it when the trendline touches the x-axis
```
aikewoody, line 136: Just wondering, does this mean that when the min value is `0` we now set it to `-1`? Because if so, when the min value is `1` it's set to `0`

selvach: Correct. This means you get this weird negative section in the graph! (See screenshot)

selvach: Support queries complaining about this are high. There must be a better fix......

aikewoody: Sorry for the late reply, this is better solution (prevents the graph from touching and x-axis and it prevents the negative section): `min: Math.max(1, Math.min(...values)),`

PR #1548, opened 2021-07-28
Revert checking of DebugScreen

selvach:
Whoops :)


bionara (approved):
Defo good to have an env-based check on this to make sure it never goes to prod too.

PR #1553, opened 2021-07-29
Restore required metro config for Android Builds

selvach:
Restore the metro config with maxWorkers 2 to ensure the build works in CiricleCI.

This was causing build failures before:


metro.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = (async () => {
+  return {
+    maxWorkers: 2,
```
aikewoody, line 3: Interesting that this is the fix, do we have any idea why we need more workers? I am asking because a clean react native project from expo doesn't have a local metro config included

aikewoody (approved):

metro.config.js
selvach: I found this:
https://github.com/facebook/react-native/issues/24716#issuecomment-492458320

PR #1563, opened 2021-08-03
Rapid test field is set by mechanism

juliatan:
# Description

Noticed a bug whereby is_rapid_test was constantly being overwritten by the wrong value, despite the mechanism being set correctly as 'lateral_flow'. https://www.notion.so/joinzoe/Updating-existing-test-to-lateral-flow-sets-it-as-PCR-instead-5020adb4f1944f929060dc6dda58b54f

This PR fixes it in that it shouldn't be overwritten by the DTO creation since this field is no longer shown to a user logging a v2 test.

TBD what we want to do if they update a v1 test (hence why I've just commented out the code for now). Let's see what comes out of wider group QA to see what users expect.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [x] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)


selvach:
Can you add a comment or just delete the line of code completely.

aikewoody (changes requested):
Better to delete that line :)

bionara (approved):

juliatan:
@selvach @aikewoody After a break and more testing, I actually decided it was better to keep it in, but move it above the creation of the `mechanism` DTO (that overwrites `is_rapid_test` field if needed for PCR / lateral flow tests). This has the benefit of allowing for the update of this field for "other" or "antibody" tests which doesn't assume anything for `is_rapid_test`.

PR #1640, opened 2021-08-31
Rename props to formikProps and improved the onSubmit functions.

aikewoody:
# Description

**Will be merged after the vaccines are live!**

- Renamed `props` to `formikProps` to make it clear the props don't come from the components render function but from the formik form render function.
- Extract as many `initialValues` definitions as possible outside of the render function as this is inefficient code.
- Renamed `formData` to `values` as this is the official term used by Formik.
- Cleaned up the `onSubmit` functions. It's painful to your eyes having code like this:
```
onSubmit={(values: TData) => {
  return this.onSubmit(values);
}}
```
Which can be shortened to:
`onSubmit={this.onSubmit}`

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)


src/features/covid-tests/CovidTestDetailScreen.tsx
```diff
@@ -190,54 +190,43 @@ export default function CovidTestDetailScreen(props: TCovidProps) {
...
-        {(props) => {
+        {(formikProps) => {
           return (
             <Form hasRequiredFields>
               <CovidTestMechanismQuestion
-                formikProps={props as unknown as FormikProps<ICovidTestMechanismData>}
+                formikProps={formikProps as FormikProps<ICovidTestMechanismData>}
                 test={test}
               />
-              <CovidTestDateQuestion formikProps={props as unknown as FormikProps<ICovidTestDateData>} test={test} />
```
juliatan, line 205: @aikewoody If `as unknown` is removed, I get Typescript errors because I think the `formikProps` already have a custom type. What's the reason for your deletion?

aikewoody: Because it's an ugly hack to work around the typescript errors you're seeing right now. Actually in a lot of places in the codebase this specific error is present and it should all be fixed in one go. I was afraid that this would be overlooked in the future as it was technically not an error

juliatan: O right, ok thanks.

juliatan (dismissed):

juliatan (approved):

PR #1713, opened 2021-09-30
Update xcode version

juliatan:
# Description

Notion: https://www.notion.so/joinzoe/Product-Eng-Board-ca9e8b1b8926481f8c94cb01a88d567d?p=7c18ad43baec44299c92490511f5f768

- The background to this ticket is essentially that I noticed I could no longer open the staging app after updating my iPhone to iOS 15.
- Based on [this](https://developer.apple.com/forums/thread/682775), a solution might be to update the XCode version that's generating the binary to v 12.5.1. 
- I'm not 100% sure if this version will suffice, or whether we need to update to v 13. Based on my local dev environment, I needed to update XCode to v 13. According to Apple notes, v 12.5.1 also doesn't explicitly says it supports iOS 15.
- CircleCI offers XCode v 13 in beta. We could try to use this version directly, but might be dangerous as their beta policy is to just keep patching the beta until it becomes stable i.e. we won't know if they introduce breaking changes.

Reference links: 
- https://discuss.circleci.com/t/xcode-12-5-1-released/40490
- https://discuss.circleci.com/t/xcode-13-rc-released/41256

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


aikewoody (approved):
Looking at the discussion on Apple forums it looks like for some people upgrading to `12.5.1` is the solution and for others it's not. I would say it's still worth trying this!

PR #1719, opened 2021-10-05
Update THR number instructions image

juliatan:
# Description

Replace THR number image following Sol's request.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 11 - 2021-10-05 at 12 15 28

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


bionara (approved):
nice one! approved - but - how does OTA perform re. caching etc when we change an image or file?would it be safer to add the change as a new name or is that just paranoia?

aikewoody (approved):

selvach:
It's safe. All the assets are bundled into a new zip file which is pushed as part of the OTA. The name can be the same.

bionara (approved):

PR #1756, opened 2021-10-19
APP BUILD 2.7.0: Upgrade React Native to v0.65

juliatan:
# Description

Related to this: https://www.notion.so/joinzoe/Explore-android-update-5cccf62daf8d43b6885b1ca41c8cf65a

### WHAT
- Upgraded to React Native to v0.65. Had to upgrade Expo SDK to ensure compatibility.
-  v0.66 was released on 01 Oct. Tried to upgrade to v0.66 but Jest was failing. This is because we couple our Jest test environment with the Expo SDK, and Expo's latest version doesn't yet support v0.66.

### WHY
- Needed minimum Android API version of 30 for Google Play Store by 1 Nov.

### IMPORTANT
- Had to remove the libraries added to `app/build.gradle` for our Android GIFs. @aikewoody We should chat about how to replace this.
- The warnings we were seeing previously on not using nested `VirtualizedLists` has now been upgraded to errors that cannot be suppressed. I did a hacky fix for the CovidTestsTabbedList and similarly for vaccines. This is not ideal and we should think about whether to refactor our base `Screen` component to remove the `ScrollView`. The alternative is to stick to this hack, as the "only" other 2 places we use `FlatLists` are in the anniversary component and disease preferences list. (See screenshot below for error.)
- `expo-permissions` which we use for push notifications has been deprecated. We need to revisit how we do push notifications as this library will be removed in the next Expo SDK upgrade.
- Had to install `fbjs-scripts` as a library to make Jest tests work. This has been fixed in React Native v0.66.
- As you might expect, there are lots more warnings about deprecations. **We should all therefore fully test this properly and make a note of the main ones to tackle as a priority**.

### OTHER
- We should think about activating React Native's Hermes (see the React Native link below for more info)? It purports to increase load times by doing smart things with the JS bundle.

### TROUBLESHOOTING TIPS
- For Detox, ensure you've got a new iOS build to make it work.
- If your iOS build passes but fails to launch because of a 405 error, go to your Mac's user folder > library > developer > xcode > deriveddata and delete all contents. Do a `yarn reset` and delete your `Podfile.lock`. Run `yarn install`. Amend the yarn ios command to set an exact simulator version, for a device you last used to run the app e.g. 
```
"ios": "yarn generate:git-hash && react-native run-ios --simulator='iPhone 11 (iOS 14.5)' --scheme 'Covid Debug' --configuration Debug"
```
- Open that version in your simulator, then run your build command again.

- If your ios build fails because of a Folly error, go to the Podfile and uncomment out the line I mentioned in that file. Delete `Podfile.lock`, run `yarn reset`, then `yarn`, and run the build command again.

### LINKS
- https://reactnative.dev/blog/2021/08/17/version-065
- https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0650
- https://blog.expo.dev/expo-sdk-42-579aee2348b6
- https://github.com/expo/expo/blob/master/CHANGELOG.md#4100--2021-03-10
- https://react-native-community.github.io/upgrade-helper/?from=0.63.3&to=0.65.0

**Screenshot**

 Simulator Screen Shot - iPhone 11 (iOS 14 5) - 2021-10-19 at 18 58 52 

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device


aikewoody (approved):
Small change, looks good to me!

bionara (approved):

PR #1792, opened 2021-11-01
Studies hub: Tab navigator with overlay

juliatan:
# Description

Notion tickets:
- https://www.notion.so/joinzoe/Studies-hub-Tabbed-nav-f137985f50ba4c91b949929d350c252c
- https://www.notion.so/joinzoe/Studies-hub-first-visit-overlay-d22f9f64d3b14970be4d1a9dc7ac9192

Includes:
- Jest tests for rendering tab navigator and overlay
- Analytics logging when tab onboarding overlay is shown
- Tested the positioning of overlay works on a range of Android and iOS devices (though the more testing the better!) 

Note:
- Had to replace `Dashboard` with `DashboardUK`, so that the flow goes from `SplashScreen` -> `Dashboard` (which now renders the `TabNavigator`). This ensures `TabNavigator` only renders AFTER the `SplashScreen`, and only for UK locale.
- There's some Android specific issues to overcome because of how Android renders child components (basically not possible to render a full screen overlay from the TabNavigator component). I therefore used React.context to solve the issue. It's a bit convoluted, but allows for the behaviour we want. Decided against passing down as props because (i) this is more flexible, (ii) with the coordinator pattern, it's complicated to pass down the navigation props and these new props whilst going through the bottom tabs navigator `react-navigation` provides.

Still to do:
- Implement close button (although this doesn't seem high priority to me).
- Can we extract overlay into its own component? It's not an easy one.
- Detox tests

- [x] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots
Note:
- Looks like 2 clicks in Android but it's just because of my GIF program.
- Android circle a little cut off because of how tab appears... I think it looks nicer like this rather than decreasing size of white circle to fit.

| iPhone X ios 12 | iPhone 13 ios 15 | 
| - |  - |
| ios-tab | ios15-tab |

| Android Galaxy Nexus | Pixel 3 |
| - |  - |
| android-tab | androidpixel-tab |

## Checklist (for submission)

- [x] Analytics/tracking events have been added (if relevant)

## Checklist (for reviewers)

## Checklist (for merger)


aikewoody:
Setting the base branch to `studies-hub/tab-nav` will make it easier to see which changes are specific for this ticket.

juliatan:
> Setting the base branch to `studies-hub/tab-nav` will make it easier to see which changes are specific for this ticket 😄

I had to redo how tab navigation was done because I realised half way through doing the overlay that it was in the wrong place and needed more rejigging.  `studies-hub/tab-nav` is therefore redundant now - this branch will include the rejigged tab nav and overlay (apologies - ideally would've been 2 branches but was too far in when I realised).

aikewoody:
@juliatan Looks like a lot of work, nicely done! About the `OnboardingContext`, you've basically reimplemented Redux on a small scale like this. Which is impressive but better to stick with one solution and keep using Redux

src/features/tab-navigation/TabNavigator.tsx
```diff
@@ -0,0 +1,250 @@
...
+    <OnboardingContext.Provider value={onboardingContext}>
+      <Tab.Navigator
+        initialRouteName="Home"
+        screenOptions={{
+          ...tabNavigatorScreenOptions,
+          ...(Platform.OS === 'android' ? tabNavigatorScreenOptionsAndroidOnly : null),
+        }}
+      >
+        <Tab.Screen
+          component={DashboardUKScreen}
```
aikewoody, line 201: Instead of using Context I think it might also be possible to pass through the function like this:
component={(props) => <DashboardUKScreen {...props} closeOnboarding={closeOnboarding} />}

PR #1893, opened 2022-01-07
Health profile: question cards

juliatan:
# Description

https://www.notion.so/joinzoe/Create-individual-question-cards-with-different-field-types-57cd4e9da4704503b2bc2adf7f24ed04

https://www.figma.com/file/7TEcX7f6rw3Z0WYR4TpX3R/ZOE-COVID-Study-(App)---MASTER?node-id=16554%3A145980

This PR creates question card components for:
- YesNo fields
- Radio button fields
- Multiple select fields (i.e. checkbox lists)
- Generic text input fields

I chose to create new components for the various field types and placed them in the `inputsRedesigned` components folder. The idea is that these would replace the existing form field components across the rest of the app over time. We would then delete the old components.

There are certain base components (`RadioButton` and `SelectableButton`) where I added a `redesign` prop to the existing component instead, rather than making whole new components. It seemed a bit overkill in these 2 cases.

The screenshots show examples of what these look like in editing / non-editing states.

NOTE: 
- This PR only deals with the design, not the functionality (which is still to be completed).
- Depending on how we convert the questions to JSON, to then render these components, we'll probably need to amend some of the logic of how it works, but the design shouldn't have to change.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

Simulator Screen Shot - iPhone 13 - 2022-01-07 at 15 18 01
Simulator Screen Shot - iPhone 13 - 2022-01-07 at 15 18 42

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


src/components/typography/TextWithBoldInserts.tsx
```diff
@@ -1,9 +1,11 @@
-import { IProps as ITextProps, Text } from '@covid/components/typography';
+import { Text } from '@covid/components/typography';
+import { IProps as ITextProps } from '@covid/components/typography/Text';
 import * as React from 'react';
-import { StyleSheet } from 'react-native';
+import { StyleProp, StyleSheet, ViewStyle } from 'react-native';
 
 interface IProps extends ITextProps {
   children: string;
+  style: StyleProp<ViewStyle>;
```
aikewoody, line 8: The type should be `TextStyle` as it's used for a `Text` component, the `ITextProps` type definition already has this defined: `style?: StyleProp<TextStyle>;`

aikewoody (approved):
A lot of work done!

src/components/typography/TextWithBoldInserts.tsx
juliatan: You're right - nice catch thanks.

PR #1918, opened 2022-01-25
Health Profile - Initialise initial form data from Redux

juliatan:
# Description

https://www.notion.so/joinzoe/Initialise-form-screens-to-read-from-Redux-state-cff26cbe55364de6b43bf9d1eeca0e06

This PR:
- Ensures that the initial Formik form state is populated with Redux state information.

Note:
- We still need to program the ability for the form screens to write to Redux state on each next button click.
- We should also decide when to call the dispatch function that calls backend API. There's probably a more efficient way of doing this.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


src/features/health-profile/screens/HPAboutYouScreen.tsx
```diff
@@ -114,6 +112,12 @@ const validationSchema = Yup.object().shape({
 
 export default function HPAboutYouScreen() {
   const patientId = useSelector<TRootState, string>((state) => state.user.patients[0]);
+  const dispatch = useDispatch();
+  const savedAnswers = useSelector<TRootState, THealthProfileData>((state) => state.healthProfile.answers);
+
+  React.useEffect(() => {
+    dispatch(fetchExistingHealthProfile(patientId)); // TODO: Should we move this to onAppStartup? Otherwise would be called on every screen change.
+  }, []);
```
aikewoody, line 120: Looks fine to me, the `fetchExistingHealthProfile` action is only dispatched once because the `useEffect` has an empty array. Also when the user navigates back from `HPFormScreen` to `HPAboutYouScreen` the use effect hook won't be executed again as react navigation is not creating a new instance of the about you screen

src/features/health-profile/screens/HPFormScreen.tsx
```diff
@@ -26,23 +28,22 @@ import { RouteProp } from '@react-navigation/native';
...
-type TFormData = { [key in THealthProfilePropertyName]: string | boolean | null };
-
 export default function HPFormScreen({ route }: TProps) {
   const patientId = useSelector<TRootState, string>((state) => state.user.patients[0]);
+  const dispatch = useDispatch();
+  const savedAnswers = useSelector<TRootState, THealthProfileData>((state) => state.healthProfile.answers);
+
+  React.useEffect(() => {
+    dispatch(fetchExistingHealthProfile(patientId)); // TODO: Should we move this to onAppStartup? Otherwise would be calling this on every screen change.
+  }, []);
```
aikewoody, line 45: Wouldn't it be enough to only have the `fetchExistingHealthProfile` action on the about you screen? Users always begin at the about you screen right? Or isn't that the case when a user presses `Save & exit`, comes back at a later time and clicks on the health profile card?

aikewoody (approved):
Great work on this!

src/features/health-profile/screens/HPAboutYouScreen.tsx
juliatan: Thanks Aike - I'm changing this to only call backend API if Redux state is empty, following our discussion yesterday.

src/features/health-profile/screens/HPFormScreen.tsx
juliatan: Exactly.. user will be taken back to the screen to last completed. I will anyway change this per my comment above.

PR #1949, opened 2022-02-16
Null values returned from backend to be converted to empty string

juliatan:
# Description

On the Health Profile, if existing patient profile values from backend are being returned as null, this causes a Yup validation error with Formik (since it needs to be at least an empty string). 

Though it shouldn't impact any specific fields anymore following PR #1947 , I've included this for robustness and as a reminder in case we need to replicate this for future forms.

- [ ] Android - Emulator
- [ ] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


bionara (approved):
Excellent! That one was a sneaky one to track down. Good hunting!

PR #1964, opened 2022-02-22
Validate the postcode for UK users.

aikewoody:
# Description

Notion ticket: https://www.notion.so/joinzoe/No-space-in-profile-Postcode-causes-map-not-to-show-up-f8fcdf4bf3a143ee8f82742697564e80

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


src/features/health-profile/utils/healthProfileUtils.ts
```diff
@@ -65,7 +67,11 @@ export const validationSchemaHPAboutYou = Yup.object().shape({
   limited_activity: Yup.string().required(),
   mobility_aid: Yup.string().required(),
   needs_help: Yup.string().required(),
-  postcode: Yup.string().required(i18n.t('required-postcode')).max(8, i18n.t('postcode-too-long')),
+  postcode: isGBCountry()
```
juliatan, line 70: Not really needed since health profile is only for UK users... but I guess no harm.

src/core/validation.ts
```diff
@@ -0,0 +1,12 @@
+class Validation {
+  regexPostalCodeUK = /^[A-Z]{1,2}[0-9][0-9A-Z]?\s[0-9][A-Z]{2}$/i;
```
juliatan, line 2: Do you know if it includes all of UK i.e. also wales, ireland and scotland?

juliatan (approved):

src/core/validation.ts
aikewoody: That's a very good point you're raising, will have to investigate into that (UK postcodes are all new to me haha)

juliatan: It might make more sense to check what the postcode / chart API allows. It might be too strict to force users to enter an exact postcode as they might not feel comfortable doing so. Let's discuss when you're next in.

bionara: This could perhaps be best used as a prompt vs validation. Ie "Your postcode appears to be invalid - are you sure it's corrct?"

PR #1977, opened 2022-03-02
Removed the sentry log on health profile submit

bionara:
# Description
The sentry log on submit is no longer needed as we have exited debug stage (ie at 100% rollout).

This correlates to a heavy traffic sentry error:
image


selvach:
Thanks for removing this. It should never have gone in.

bionara:
> Thanks for removing this. It should never have gone in.

It was required for 1 and 2 % rollout to help debug, but should not have gone in to 100%. Or more aptly - when rolling out to 100%, we should have done an OTA first to remove this.

Going forward, we may want to have some startup_info flag to safety check these, ie error logging while under a certain threshold.

juliatan:
@bionara Let's also delete the Logging action in the `saveExitOnPress` function within `HPFormScreen.tsx`. We can then also delete `EHealthProfileContext` and `health_profile` from `TFeatureName` in `Logging.ts`.

juliatan (approved):

PR #2000, opened 2022-03-23
Refactor Amplitude tracking

juliatan:
- Removes all unused events
- Removes automated tracking of all screen views and implements opt-in tracking of specific screens instead


imanmzoe (approved):
interesting we can automate this when we shift to github actions to open a PR on merge to develop :)

juliatan:
> Details

Yes definitely! We also have to manually add new releases and tags, so want to look at how we automate this with GH Actions also. :)

PR #2003, opened 2022-03-24
Cut down list of screens being tracked (#2002)

juliatan:
- Further reduction in screens being tracked
- Fix bug where new users were being taken back to profile sign up screen, when they went to log a test or vaccine from the profile tab


ThomasEnglish (approved):
Looks good to me.

Ran through the registration process, and reporting tests/vaccines through profile with no issue.

PR #2004, opened 2022-03-24
Fix bug for new user signups and profile tab

juliatan:
# Description

https://www.notion.so/joinzoe/Bug-New-user-signup-going-via-profile-tab-to-add-tests-vaccines-results-in-user-being-taken-back--7a4f0bd61df248fa972038c4dedf984e

https://www.notion.so/joinzoe/Bug-Done-button-on-vaccines-screen-via-profile-tab-doesn-t-work-if-there-are-no-vaccines-logged-e77c2fcf174d4f1bb6549ad27a1895ee

New users who sign up and immediately go to log a vaccine and test via the profile tab were being shown the "fill in your profile" screen instead. This is because the instance of the patientCoordinator was not being updated in time, upon the new user signup flow being submitted to the backend.

To get around this for now, the COVID test and vaccine card links from the profile tab initiates a different function on the assessment flow coordinator instead. This is already noted under technical debt, but what we want to do in the future is to break from this coordinator pattern and single class instance, which will make this ability to navigate from different screens more flexible.

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


ThomasEnglish (approved):
Looks good to me. No obvious issues standing out, but would like to do some testing on the staging environment to ensure flows work for existing patients 👍

PR #2009, opened 2022-03-28
NO-YT: Add storybook

imanmzoe:
# Description

We want to add storybook to our tool belt to supercharge our Design to Engineering workflow. This tool will allow us to easily add/change/QA new components. To be able iterate as fast as possible we need to ideally spend as little time as possible on base level components or minimal changes due to not having the right level of flexibility to begin with. Following the atomic design system we will slowlt transition all current base components into 3 layers - atoms, molecules and organisms the fill in any missing components as needed.

## Screenshots

## Checklist (for submission)

- [ ] Add storybook libraries
- [ ] Update Android binary
- [ ] Update App startup to use new config variable

## Checklist (for reviewers)

- [ ] Checked app can be run without storybook as usual
- [ ] Checked app can be run with storybook + web UI


package.json
```diff
@@ -102,7 +106,7 @@
     "react-dom": "16.13.1",
     "react-native": "0.65.1",
     "react-native-calendar-picker": "7.1.2",
-    "react-native-config": "1.4.4",
+    "react-native-config": "^1.4.5",
```
juliatan, line 109: We tend to lock the specific version of the packages i.e. removing the `^`. This was due to some errors seen in the past that was really difficult to debug but was ultimately found to be caused by unnoticed package updates.

juliatan (commented):
Exciting stuff! Just made a small comment.

imanmzoe:
closing this PR, will replace with a newer one

PR #2012, opened 2022-04-01
Display LFT Prompt

juliatan:
# Description

https://www.notion.so/joinzoe/LFT-Prompt-Modal-for-newly-sick-d46770dabc144f2fb318094b4362b3b8
https://www.figma.com/file/7TEcX7f6rw3Z0WYR4TpX3R/ZOE-COVID-Study-(App)---MASTER?node-id=18701%3A168826

How the feature works:
- Backend sends through 2 flags (one of the initial modal and one for the reminder modal) on app start up that checks, based on historical COVID tests and assessments logged, whether a patient should be shown the LFT prompt.
- The frontend (on the ThankYou screen) goes through additional checks to see whether the patient has logged a test today and logged COVID symptoms, before showing the modal.
- We store a record of how many times we've shown the modal and date it was seen in Redux. This allows us to display the reminder modal at the appropriate time (assuming the backend flag returns true).

- [ ] Android - Emulator
- [x] iOS - Emulator
- [ ] iOS - Physical device

## Screenshots

simulator_screenshot_D9926E0B-0D54-47FF-A320-C91A38B78F54

## Checklist (for submission)

## Checklist (for reviewers)

## Checklist (for merger)


src/core/state/covid-tests/types.ts
```diff
@@ -0,0 +1,27 @@
...
+type TTestLoggedToday = TTestLoggedTodayBase & TTestLoggedTodayConditional;
+
+export type TStateCovidTests = { lftPrompt: { [patientId: string]: TTestLoggedToday } };
+
+export type TUpdateDateLFTPromptAction = {
+  patientId: string;
+  dateLastSeenLFTPrompt: Date;
+};
+
+export type TUpdateCurrentPatientLoggedToday = {
```
imanmzoe, line 24:
```suggestion
export type TUpdateCurrentPatientLoggedTodayAction = {
```

to stick to the convention for Actions?

src/core/state/covid-tests/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type TTestLoggedTodayConditional =
+  | {
+      dateLastSeenLFTPrompt?: Date;
+      timesSeenLFTPrompt: 0;
+    }
+  | {
+      dateLastSeenLFTPrompt: Date;
+      timesSeenLFTPrompt: Exclude<number, 0>;
+    };
```
imanmzoe, line 13: Hmm 🤔  I'm actually not familiar with this syntax of two dictionaries separated by `|` which I think means the regular use of `or` but for two different dicts altogether but for my own learning what is the difference between the above and the below:

```
type TTestLoggedTodayConditional =
   {
      dateLastSeenLFTPrompt?: Date;
      timesSeenLFTPrompt?: 0 | Exclude<number, 0>;
   }
```

src/features/covid-tests/screens/CovidTestListScreen.tsx
```diff
@@ -42,6 +44,14 @@ const getTodaysLFT = (covidTests: TCovidTest[]) => {
   return todaysTest;
 };
 
+const checkIfPatientLoggedPCRLFTToday = (covidTests: TCovidTest[]) => {
+  const todaysTest = covidTests.find(
```
imanmzoe, line 48:
```suggestion
  const loggedLFCPCRTestToday = covidTests.find(
```

src/features/covid-tests/screens/CovidTestListScreen.tsx
```diff
@@ -42,6 +44,14 @@ const getTodaysLFT = (covidTests: TCovidTest[]) => {
...
+    (test) =>
+      test.mechanism === ECovidTestMechanismOptions.NOSE_OR_THROAT_SWAB && dateIsToday(test.date_taken_specific),
+  );
+  return !!todaysTest;
```
imanmzoe, line 52:
```suggestion
  return !!loggedLFCPCRTestToday;
```

src/features/thank-you/ThankYouUKScreen.tsx
```diff
@@ -13,43 +16,104 @@ import { selectStartupInfo } from '@covid/core/state/selectors';
...
+  'fatigue',
+  'sore_throat',
+  'headache',
+  'sneezing',
+  'persistent_cough',
+  'hoarse_voice',
+  'chills_or_shivers',
+  'unusual_joint_pains',
+  'loss_of_smell',
+  'fever',
```
imanmzoe, line 54: minor but should these be enums?

src/features/thank-you/ThankYouUKScreen.tsx
```diff
@@ -13,43 +16,104 @@ import { selectStartupInfo } from '@covid/core/state/selectors';
...
+  const currentAssessment = assessmentService.getCurrentAssessment();
+  const covidTestsState = useSelector<TRootState, TStateCovidTests>((state) => state.covidTests);
+  const currentPatientId = currentAssessment.patient || '';
+  const lftPromptStateForCurrentPatient = covidTestsState?.lftPrompt[currentPatientId];
+
   const [askForRating, setAskForRating] = React.useState<boolean>(false);
   const [showHealthProfileReminderModal, setShowHealthProfileReminderModal] = React.useState<boolean>(false);
+  const [showLFTPromptModal, setShowLFTPromptModal] = React.useState<boolean>(false);
 
+  const fiveDaysAgo = moment().add(moment().utcOffset(), 'minutes').subtract(5, 'days').toDate();
```
imanmzoe, line 83: just making a note here, but the moment library is actually not maintained anymore sadly and we should probably use the suggested alternative on their repo (which I believe is the native one, not sure)

imanmzoe (commented):
some really minor comments, otherwise, nicely done! 

❓ not related but just in general, we don't use type hints for the simple / obvious ones but I was wondering why, like for `loggedLFCPCRTestToday`

Is it just a style thing?

src/core/state/covid-tests/types.ts
juliatan: Yup, so the `|` is just a regular union type, but what this is saying, when combined with `TTestLoggedTodayBase` in line 15, is that if `timesSeenLFTPrompt === 0`, then `dateLastSeenLFTPrompt` can be `undefined`. If `timesSeenLFTPrompt !== 0`, then `dateLastSeenLFTPrompt` MUST be a `Date`.

I wanted to avoid the cast where Typescript complains about `dateLastSeenLFTPrompt` being `undefined`, when it's legitimately undefined. 🙂

src/features/thank-you/ThankYouUKScreen.tsx
juliatan: Yup you're right. Added a ticket to our refactoring epic: https://www.notion.so/joinzoe/Replace-momentJS-in-RN-code-base-since-it-s-no-longer-maintained-74e885a5dd9649e7b162fbff1a2d27a3

juliatan:
> some really minor comments, otherwise, nicely done!
> 
> ❓ not related but just in general, we don't use type hints for the simple / obvious ones but I was wondering why, like for `loggedLFCPCRTestToday`
> 
> Is it just a style thing?

Do you mean why we don't explicitly type the return value? If so, I generally prefer to have Typescript infer the types where it can (since it makes it more flexible e.g. in cases of refactoring). It's only if Typescript complains about a value potentially being `any` that I then explicitly type it.

src/core/state/covid-tests/types.ts
imanmzoe: ahhh makes sense 👍

